Issue #1, opened 2022-10-12
Incompatible with gnome shell 42


Klewerro:
Plugin at [gnome.extentions.org](https://extensions.gnome.org/extension/4928/mpris-label/) is incompatible with Gnome Shell 42 (Manjaro, shell version 42.4).
I forced installing incompatible extensions by command `gsettings set org.gnome.shell disable-extension-version-validation true`, and after few days of using plugin, I didn't noticed any bugs or issues.

Please consider test and raise up compatibility version, because it's very useful and straight forward plugin :+1:

Batwam:
works fine with Gnome 43 by disabling the version check.

Moon-0xff:
Good to know, thanks to both of you for testing my extension. Gonna bump the version soon!

Batwam:
updated compatibility included in pull request #3 for good measure

Moon-0xff:
Uploaded the extension with explicit support for 42 and 43 to extensions.gnome.org
I'm going to close this thread when the "new" version gets accepted.

Moon-0xff:
Looks like the "new" version got accepted, Closing this thread.

PR #3, opened 2022-10-30
Version 4 Development

Batwam:
added button to turn on/off in settings. checks the status and hides the text if option is set and playback paused.


Moon-0xff:
Your code seems to be very on point, however it breaks the control flow of the program.

Some statements are called way too early, breaking some other functionality, specifically:
`When the current playback is paused, you can't switch to another one`
and possibly there's some other hidden bug there.

There's some other nitpicks to do, but in any case this is a great start! Thanks for your help!

extension.js
```diff
@@ -8,7 +8,7 @@ const ExtensionUtils = imports.misc.extensionUtils;
 let LEFT_PADDING,RIGHT_PADDING,MAX_STRING_LENGTH,EXTENSION_INDEX,
 	EXTENSION_PLACE,REFRESH_RATE,BUTTON_PLACEHOLDER,
 	REMOVE_REMASTER_TEXT,DIVIDER_STRING,FIRST_FIELD,SECOND_FIELD,
-	LAST_FIELD;
+	LAST_FIELD,REMOVE_TEXT_PAUSED;
```
Moon-0xff, line 11: I prefer the name `REMOVE_TEXT_WHEN_PAUSED`. a little bit more descriptive

extension.js
```diff
@@ -116,6 +123,7 @@ class MprisLabel extends PanelMenu.Button {
 		FIRST_FIELD = this.settings.get_string('first-field');
 		SECOND_FIELD = this.settings.get_string('second-field');
 		LAST_FIELD = this.settings.get_string('last-field');
+		REMOVE_TEXT_PAUSED = this.settings.get_boolean('remove-text-paused');
```
Moon-0xff, line 126: Same as above

extension.js
```diff
@@ -138,7 +146,13 @@ class MprisLabel extends PanelMenu.Button {
 			if(!this.playerList.includes(this.player.address))
 				this.player.changeAddress(this.playerList[0]);
 
-			this.buttonText.set_text(this._buildLabel());
+            this.status= new Status(this.playerList[0]);
```
Moon-0xff, line 149: Move this to _buildLabel

extension.js
```diff
@@ -204,6 +217,23 @@ class Player {
 	}
 }
 
+class Status {
```
Moon-0xff, line 221: Rename to getStatus for consistency

extension.js
```diff
@@ -219,7 +249,7 @@ function getPlayerList () {
 }
 
 function parseMetadataField(data) {
-
+//log('hello2');
```
Moon-0xff: Missed it!

extension.js
```diff
@@ -251,7 +281,10 @@ function parseMetadataField(data) {
 function removeRemasterText(datastring) {
 	if(!REMOVE_REMASTER_TEXT)
 		return datastring
-
+
```
Moon-0xff, line 284: Move this to the very top of `_buildLabel`

This should return an empty string if there's only one player, `BUTTON_PLACEHOLDER` if there's more

Moon-0xff: `this.playerList` is the list of mpris sources

Moon-0xff (commented):
Here's a quick review, if you want to contribute further

extension.js
Moon-0xff: Upon further inspection this statement looks a bit useless or a bit of a bodge.
If the player is paused and `REMOVE_TEXT_PAUSED` is true this function shouldn't be called, but that was my mistake

extension.js
Batwam: yeah, I called it status for consistency with "Player". Ideally, I would have liked to combine the Metadata and PlaybackStatus checks but couldn't seem to get it to work without essentially duplicating the whole Player class...

extension.js
Batwam: will do

extension.js
Batwam: will do

extension.js
Batwam: Hang on, what’s the idea behind `BUTTON_PLACEHOLDER` again? Seems to be linked to having multiple players but I just implemented a check to try identify the “Playing” software so not sure if the new method is consistent.

Moon-0xff: `BUTTON_PLACEHOLDER` is there because some applications "steal" the focus of the extension with an empty string. The placeholder string should hint the user to switch to another mpris source if they want.

By the way, i have not tested your new changes yet.

Moon-0xff: Tested your changes. I see two problems:
* When another source starts to play, the extension jumps to the new one, this could be particularly annoying if you are browsing reddit or whatever, although some people will probably prefer this behaviour (new option?)
* When two or more sources are playing simultaneously, you can't switch between them

Moon-0xff:
Refrain from refactoring your code any further, i'm going to push changes to your branch. Alternatively i can merge the pull request right away. It's your decision.

Moon-0xff:
That's my work for the day. Last commit breaks the extension.
Try to debug my stuff if you wish! Thanks again.

Moon-0xff:
Added a very basic "install script" for faster development, only execute it from inside the directory.
Use `GNOME-xorg` and reset the session with `Alt+F2 -> r` for better effects.

extension.js
Batwam: - I wasn't aware of this case. It would indeed switch as the code is currently iterating from last to first. We could iterate from first to last as I assume that the browser would be added at the end of the list and this would then be consistent with current behaviour?
- Yes, when two apps are playing at the same time, the tool won't know which one to pick but this is already the case. I am not sure how often people listen to multiple songs at the same time though. What would be the expected behaviour if multiple sources are playing, I assume that you would need to pick one as I doubt you would want to list them all?

Batwam:
> Refrain from refactoring your code any further, i'm going to push changes to your branch. Alternatively i can merge the pull request right away. It's your decision.

feel free to merge right away if you are happy with the changes

extension.js
Batwam: ok, upon further review, the _buildLabel() function no longer , upon further review, the _buildLabel() function no longer  hides the text when playback is paused because of the conditions are only met if there are multiple players and `labelstring` is empty which wouldn't normally be the case. 

The reason for cycling through the players to find out if one was actively Playing was so the text could switch automatically to the active player when one is paused and the other one is started. With the current code, this check isn't getting executed as it only runs if `(labelstring.length == 0)`. Arguably this is a separate suggestion to the original "hide text when playback is paused" so we can certainly discuss it separately if needed. I think that now that you have the ability to figure out if an app is playing, you could use this to skip the player returning empty strings and return the info for an actively playing app `this.player = new Player(this.playerList[i])` with non-zero info (I hope this makes sense)? This way you can try to switch automatically and only return `BUTTON_PLACEHOLDER` in edge cases where there is only one player and it is returning an empty string?

I have an updated code which I think goes some way towards adressing the above. It iterates from 0 so hopefully reddit won't be stealing the focus if started after the music player. if you are worried it might still be and issue, we could easily include the string length check within the loop to exclude players returning empty `labelstring`. Let me know if this is of interest and I can push it as commit.

Moon-0xff:
I'm going to respond to your last comment here.
> I wasn't aware of this case. It would indeed switch as the code is currently iterating from last to first. We could iterate from first to last as I assume that the browser would be added at the end of the list and this would then be consistent with current behaviour?

Not relevant anymore. Current implementation (same as the old one) only iterates `playerList` when selected mpris source no longer exists.

> Yes, when two apps are playing at the same time, the tool won't know which one to pick but this is already the case. I am not sure how often people listen to multiple songs at the same time though.

People with faulty or badly designed amplifiers/stereos/headphones often play a "silent song" to prevent their devices to turn themselves off or buzzing. Also the aforementioned 'browsing reddit or whatever while playing music' scenario counts as two distinct mpris sources.

> What would be the expected behaviour if multiple sources are playing, I assume that you would need to pick one as I doubt you would want to list them all?

You need to pick one.

Moon-0xff:
We are done, i believe.
But now, we could implement your unintended feature as an option:
> When another source starts to play, the extension jumps to the new one, this could be particularly annoying if you are browsing reddit or whatever, although some people will probably prefer this behaviour (new option?)

Using `playerList` and the new `activePlayers` list, it should be easy.

extension.js
Moon-0xff: i don't know if this applies to the present version of the file. Check and test the last commit and if you believe your implementation it's better, push it. We can always nitpick or revert the changes later.

Moon-0xff:
Made a quick implementation, it works, but you can't switch sources with the option enabled.
Gonna finish it tomorrow, unless you beat me to it.

Batwam:
Hey, thanks for this. I'll let you complete the updates if that's ok as I don't want to mess it up and will check how it's implemented it once it's done but it's looking good so far with the new `activePlayers` list filtering out the non-playing sources :-)

Batwam:
ok, I've fixed up an issue with the `AUTO_SWITCH_TO_MOST_RECENT` section returning an error if all players are paused which I believe you had picked up on already. I check with the two new options on/off and it all appears to work as expected.

Moon-0xff:
No, i was not aware of this. Gonna keep it in mind.
Better yet, i'm going to look for another way of getting a last index

Moon-0xff:
It looks like Array.prototype.at() works in gjs
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at

Moon-0xff:
Yesterday, trying to implement the auto switch option did more or less the exact changes. Plus a really complicated and sloppy way of sorting `activePlayers` by activity that didn't work and sprawled itself to every corner of `extension.js`

The commits of today are the working parts of it. Plus some improvements.

I wasn't able to come up with a good way of sorting `activePlayers` correctly. `getActivePlayers` returns the list sorted by spawn order, meaning the time when the application which contains the mpris interface was opened.

I want this auto switch option to jump to the most recent instance of a player going from `X` to `Playing`
Sorting a list isn't the best way of achieving that, but it's a lot more flexible.

I'm also not a fan of how i ultimately implemented the trigger of the auto switch in commit [1d6dedc](https://github.com/Moon-0xff/gnome-mpris-label/pull/3/commits/1d6dedc144079c3c2586704963427d91f30b09f9)

Moon-0xff:
This is(was) my approach:
1. Get the list of active players from `getActivePlayers` (once)
2. Detect when a player changes it's playback status to `Playing` using both `activePlayers` and `playerList` and move it to the end of `activePlayers`
3. Repeat step 2 every cycle

Batwam:
Note that I noticed that when I just start an app like Spotify and start playing musing, the text doesn't appear. However, as soon as the song changes or I switch to the next song manually, the text appears. I can't quite figure out what is happening but based on my investigations, `this.proxy.Metadata[field].get_strv()[0] is empty` is returning empty metadata when the app is just started so `parseMetadataField(data)` returns `""`. I tried with an older version of this commit I had backed up and the issue was already there so it doesn't seem to be something which has been introduced recently.

Moon-0xff:
A year ago when testing this extension Spotify returned at startup `xesam:something` and i had to manually filter it out in the code. It wouldn't surprise me if today this startup sequence of sorts it's slow enough to notice. That's my theory at least.

Moon-0xff:
Perhaps removing this from `parseMetadataField` and testing it again could reveal more information
```
    if (data.includes("xesam:") || data.includes("mpris:"))
        return ""
```

Batwam:
Just to confirm, I have only experienced the issue with Spotify. 
As far as I can tell, `data` fed to `parseMetadataField(data)` is empty so the code stops are the previous check since `data.length == 0`. I fired up D-Feed to inspect the Bus info and the Metadata does appear to be loaded so the info is there but while `getPlayerStatus(playerAddress)` is returning the updated status (shows Playing if I click start the music), `this.proxy.Metadata[field].get_strv()[0]` remains blank so something doesn't seem to get loaded up...

Moon-0xff:
Nice catch!

Moon-0xff:
I want to convert `playerList` and `activePlayers` to a single, keyed collection.

Moon-0xff:
Doing this changes to improve in all the D-bus stuff in the future

Moon-0xff:
Made a lot of changes, gonna push them tomorrow because they are still half baked.

Moon-0xff:
Well now I'm finally satisfied with the implementation. It's not the best but certainly it's very flexible.
The auto switch it's behaving exactly as i want:
> I want this auto switch option to jump to the most recent instance of a player going from X to Playing

And it doesn't seem to spam the logs with warnings or errors.
I don't trust it though. It definitely needs some serious debugging and a lot of testing.

Batwam:
Looking good so far. The only warning I noticed is `Usage of object.actor is deprecated for MprisLabel`. 
https://lazka.github.io/pgi-docs/Clutter-1.0/structs/Container.html

Batwam:
ok, I've done a partial attempt to update the depreciated calls based on best-guess for the syntax. It doesn't seem to have any adverse effect on the extension but the error is still there, possibly due to the use of `.remove_actor()`, not sure...

Moon-0xff:
Updated it completely now. Guessing work honestly.
Just changing `remove_actor` to `remove_child` goes well and changing extension place through settings doesn't throw any warnings or errors, however when disabling the extension it *sometimes* throws `this.actor is null`
Using `this.container.destroy` instead of `this.destroy` makes the error appear less often?
There's probably a better way though, perhaps `super.*something*`

Moon-0xff:
Last one looks big, but there's not a lot of changes to the *underlying program structure*

Really i should have pushed it with a extended commit message so here's one:
`_init`: Remove unnecessary lines
`_updatePlayerList`: Fix the code and condense some loops
`_pickPlayer`: Restores behaviour when no active players are present, and rename a variable for clarity
`buildLabel`: Changes for clarity

Batwam (commented):
you might want to revert this latest commit as it no longer switches to the recent source.

Moon-0xff:
This one maintains both the ability to cycle sources AND auto-switching to the most recent source, however it only deals and switches to **active** ones. Not a bad trade-off

Batwam:
See proposed update for the spam issue.

yeah, I agree with not going over the top on depreciated functions as this could have the unintended effect of breaking compatibility with older gnome versions. It it's not broken, don't fix it...

Moon-0xff:
Ha! I didn't noticed that! 
Anyways, have you been running your latest commit? Any problems so far?
Talking about problems the one about the extension throwing `this.actor is null` looks like it has nothing to do with the extension

```
JS ERROR: TypeError: this.actor is null
        _syncEnabled@resource:///org/gnome/shell/ui/windowManager.js:138:25
        onStopped@resource:///org/gnome/shell/ui/windowManager.js:150:35
        _makeEaseCallback/<@resource:///org/gnome/shell/ui/environment.js:150:22
        _easeActorProperty/<@resource:///org/gnome/shell/ui/environment.js:316:60
        _destroyWindowDone@resource:///org/gnome/shell/ui/windowManager.js:1596:21
        onStopped@resource:///org/gnome/shell/ui/windowManager.js:1564:39
        _makeEaseCallback/<@resource:///org/gnome/shell/ui/environment.js:150:22
        _easeActor/<@resource:///org/gnome/shell/ui/environment.js:239:64
```
As we were at the time fidgeting with the actor stuff i just assumed this had something to do with the recent changes to the extension. If that were the case i believe the log would be something like this:

```
JS ERROR: TypeError: this.actor is null
        disable@$HOME/.local/share/gnome-shell/extensions/mprisLabel@moon-0xff.github.com/extension.js
```

It could be that indeed it has something to do with the extension, but seems unlikely, and in that case the fault is on GNOME.

Batwam:
Hmmm ok, yeah, I did notice that it wasn't going away.

No issue on my side with the latest commit so far. All working as intended.

Moon-0xff:
Well then, we are *almost* done here.

The question about using `this.destroy`, `this.container.destroy` or something else is still open. I **believe** it doesn't matter, but in the face of ambiguity it's probably wiser to revert it to `this.destroy`

As this latest version looks good and stable, and unless you want to contribute some additional refactoring, it's time to test it with some long casual use, i believe a week would suffice.

Batwam:
Yeah, let's give it a week. I'll let you know if I encounter any issues. Feel free to update the code if needed and I'll update to the latest.

Moon-0xff:
As the commit says, based purely on my own speculation. Here's my reasoning:

* **rename function `disable` inside MprisLabel class to `_disable`**
gnome-shell (or something else i don't know) looks in `extension.js` for functions `disable` and `enable`. It can encounter them (at  least enable) inside `MprisLabel` (see commit 28b96ae and 487e187) , so *perhaps* it's safer to not repeat those names
* **Don't assign `this.player` to `null`**
shouldn't matter
* **Call `destroy` from `disable`**
Already tested it, used in most extensions, sounds safer to call `destroy` from the outside?

Been testing it a bit, no logs so far.

Batwam:
ok, no issue with the two nea options enabled. However, interesting behaviour I have noticed if "Remove text when paused" is **disabled**:
- start source #1, title appears. pause, title stays. all good
- start second media player, title appears (all good so far). pause, title goes away (it shouldn't), showing the title of the source #1

Moon-0xff:
If you have the auto-switch option enabled, **that's expected**, it's part of the auto-switch trade-off mentioned above.
Really it shouldn't, it should switch only if a **source changes state from X to Playing** (minding that a source can spawn already Playing). But i wasn't able to do it properly.

Batwam:
Yeah, I see what you mean. Rather than monitor status change, would a possible trade-off be that it would switch only if the other source is playing?

Batwam:
try this please, it's working for me. I also tried with the auto switch turned off (something I probably wouldn't use) for completeness and it seems to work fine too.

Batwam:
Also, maybe something for a future version but I'll put it out there. The new "auto switch" is a solid feature. The only reason I can think of where people may not way to activate "auto switch" or be bothered by it would be if something like their browser is taking over the text while they have the music playing and the browser on mute. Do you think that it would be worth considering an option which would do "Ignore browser media sources" and filter out mpris sources like chromium? Not something which affects me personally but I'd be interested to get your thoughts on this.

Moon-0xff:
It's on the menu, yes. A filter list for mpris sources. I'm not eager to start working on it though. Really i should write a to-do / help wanted list.

Batwam:
Yeah, please make a list, I'd be really interested to see it!

Moon-0xff:
Nice to know, going to add it as soon as I'm able.

Batwam:
ok, all good so far with the latest version. By the way, when you click on the Reset button in settings, all settings (`SpinButton`, `Entry`) are set back to defaults except the ComboBox values (`extension-place`, `first-field`,...). It looks like Combobox contents aren't handled correctly, perhaps because the values are blank so the extension doesn't know which field to select?

Moon-0xff:
Good catch, it seems comboBox values are correctly set to default, however the values selected in the comboBoxes aren't changed. Last commit addresses this. Just copy pasted a few lines.
Does this happen in GNOME 3.x too?

Batwam:
I can confirm that this works now. I have checked gnome3.x but I can fire up the old laptop when I get a chance as I believe that it is running gnome3.36.

Moon-0xff:
> ok, works fine in gnome 3.36.9.

Good to know, would be a shame if compatibility with GNOME 3.3x wasn't maintained
> If you go back and forth between the tabs, so does the extension which can feel a bit weird.

I'm not sure if that's the extension, have you checked how many chromium instances are in `playerList`?
If there's only one, it means the selection is made by chromium

> it may be useful to consider implementing something like "ignore muted sources" [...] This should be easy based on the dbus info which gives a volume of 0 for mute sources.

Good idea!

Moon-0xff:
A new option for a fancier behaviour, at least i think so.
I haven't tested it against any edge case yet.

Batwam:
quick proof of concept to filter out muted sources. I tried to filter at the `updatePlayerList` level but couldn't quite figure out how to implement the filtering so I put it further down in `_pickPlayer`. Just an idea for now, feel free to ignore. Also, it seems like some sources like chromium don't actually provide the volume info? I'm getting a value of `1` regardless of the volume the tab is set at within Youtube/Spotify. If the tab itself is muted, the source goes away anyway so this is perhaps not that useful...

Moon-0xff:
Have you tried it with another applications? Tried myself, and doesn't seem to have any effect :/

Batwam:
yeah, it works for me with the Spotify app as well as Rhythmbox (although by Default, the "Alternative Toolbar" plugin is activated, at least in Ubuntu, which hides the volume button...). With chromium, it's always reporting Volume=1 regardly of what I set Spotify/Youtube inside the tab at. 

I also installed firefox to check and it's reporting `null`... I checked with d-feet and while Volume is listed, it's not returning any value for Volume, only the values listed with a value below:
image

Moon-0xff:
Then it's good, but I'm afraid some rogue application reports *muted* when it shouldn't.
Please test and review latest commit, and be mindful of your whitespace!

Batwam:
sorry for the rough formatting, it was just a quick test I was doing to see if removing muted sources would work. On the plus side, I finally found the option to "Render Whitespaces" in the IDE, that should help!

Based on the issues encountered (no visibility of volume within tabs, [firefox not reporting volume](https://bugzilla.mozilla.org/show_bug.cgi?id=1674654), I think that we should scrap the idea and revert back to [e753801](https://github.com/Moon-0xff/gnome-mpris-label/pull/3/commits/e75380128c6c3e94f920dc384c912d9ea7278ca6), pending implementation of a filter list which would allow filtering out browsers.

Moon-0xff:
Yes, you are right. It's not a bad addition though.
Maybe it could be added back in the future. Specially if firefox and chromium begin to report their volume correctly.

Moon-0xff:
Hopefully i reverted the changes correctly and the extension it's in good shape. If that's the case, and with your approval, I'm going to merge it.

Batwam:
No problem with the latest commit. I don't think that you need my approval to merge but if you do, you have my blessing 😉

Moon-0xff:
As reassurance, you are certainly good at detecting my mistakes.

Batwam:
No worries, it's always much easier to spot other people's mistakes! Thanks for your patience fixing mine and showing better ways to integrate the changes into the code, it's a great learning opportunity 🙂

Moon-0xff:
Well, thanks to you for submitting the PR in the first place. And for all the help. code and ideas you contribute it along the way.

Before merging this i want to discuss and hear your ideas for the next release, as there's still a lot that could be improved/added. But we are delaying the merge too much already. What about you open it as an issue? Or a pull request if you have some code for it.

PR #5, opened 2022-11-25
Todo 1 - Icons alongside the label

Batwam:
Here you go, first proof of concept including an app icon next to the text. I'd a rough test and spotify is loaded statically but it can also load the firefox and google-chrome icons. This works on startup only for now, I haven't worked out how this can be refreshed yet.
image

There will be more work needed to:
- make this an option (that's the easy part)
- apply the padding differently depending on whether there is an icon it now (or simply skip the image and keep icon/label)
- filter the app name from the address. There may be some work needed to extract the source name from the adress (browsers include an instance number) and interpretation work needed. For instance, I suspect that google-chrome and chromium both report as chromium in dbus.
image
- check that it works with various installs (normal install, flatpak, snaps...) and various distros
- recognise the original project which inspired the code (sp-tray)
- formatting clean up. there might be some rogue white spaces here and there from the original code, sorry... I've also included some temporary comments to discuss what I tested.


Moon-0xff:
Wow you are fast. But [there's a lot of questions and problems regarding that code](https://github.com/esenliyim/sp-tray/issues/26)

We need to understand what we are adding to the repo if we want to maintain this extension independently. Also, as you can read in the link, we should thank @koolskateguy89 for the code i suppose.

Moon-0xff:
Sorry for reverting your commits.
Added icon.js as the place where to dump our changes. That way our commits are easier to follow.
I didn't added back all the koolskateguy stuff, only the bare minimum.

I have been trying to get the icons the same way [AppMenuButton](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/gnome-43/js/ui/panel.js#L34) does.
AppMenuButton is the label for the focused window. With default everything it's at the left of our extension.
Inspecting the code it looks like an extension, but it's async all the way down and really hard to follow.

If we can implement this as GNOME does it, then we shouldn't worry too much about missing icons as GNOME seems to handle icons just fine.

Batwam:
So we can load up an icon when the extension loads which is nice but do you have any idea how to swap to another icon after that? Once we have that, we can look to find a good location for this icon update which doesn't require loading the icon 3 times per second... Ideally this would only be done when the source is being changed.

Moon-0xff:
We can change icons by first removing it from `box`, redefining it, then adding it back. Try this:
```
                this.box.remove_child(this.icon);
                this.icon = new IconHandler().getIcon("rhythmbox");
                this.box.add_child(this.icon);
```

Moon-0xff:
Today wasn't too hard to follow [AppMenuButton](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/gnome-43/js/ui/panel.js#L34).
Take a look at the last commit, i managed to copy AppMenuButton behaviour.

Moon-0xff:
OK, I'm doing some real progress now. If all goes well I'm going to push it tomorrow.

Batwam:
Yep, that's looking good, although focus_app if of less interest to us so we'd need to find a way to get the icon for the app associated to a given mpris source address. Regarding multiple browsers having the same "chromium" source name, perhaps there is a way to get a list of active apps and pick the most likely (or actual app name from link to mpris source dbus info but I haven't see it in the properties yet).

One thing I was thinking also is rather than have the option as "Show source icon (Experimental)" `true/false`, perhaps we could consider `none/left/right` which would allow not only to turn on/off but also decide left/right? I tend to prefer left btw so as long as I can get it on the left I'm happy, it's just that I see that it's on the right lately.

Moon-0xff:
We can get the DesktopEntry directly from DBus? That's a real game changer.

Moon-0xff:
> perhaps there is a way to get a list of active apps and pick the most likely

Yes that's how i was implementing it.

We can get the list of active apps using this snippet:
```
                let appSystem = Shell.AppSystem.get_default();
                let runningApps = appSystem.get_running();
```
this returns a list of [Shell App Objects](https://gjs-docs.gnome.org/shell01~0.1_api-app/). We can get the desktop entry from this using [this](https://gjs-docs.gnome.org/gio20~2.66p/gio.desktopappinfo).
How do we know what `App` belongs to `this.player.address`? I haven't found a form of knowing for sure but we can guess it fairly accurately (i think).

Moon-0xff:
But if we have the desktop entry along with the DBus then we have the icon too!

Batwam:
Indeed. Let's hope that we can go straight from dbus info to icon but I have the feeling that not all sources will provide the info. Let's have a look after your next commit when we get a better feel for what the latest code looks like. I'm thinking about using the dbus info to match first and the best guess based on source address cross checked against active apps as a fallback if we find sources which don't provide the DesktopEntry info.

Moon-0xff:
After trying various ways of getting the icon, i settled on this way.
It's based around [this method](https://gjs-docs.gnome.org/gio20~2.66p/gio.desktopappinfo#function-search). It returns a [ThemedIcon](https://gjs-docs.gnome.org/gio20~2.66p/gio.themedicon). And that's probably better.

This last commit barely works, it's complicated, and it looks like the extension is broken in other places too. But at least demonstrates the idea well enough.

Batwam:
No worries. Yes, I can see that it loads the rhythmbox icon but not chrome or spotify (flatpak). It also no longer switch sources or remove the text when the source is gone. 

Anyway, it's running and an icon is loaded so it's a good start! Thanks for including comments in the code, that should help work out what needs to be done.

Moon-0xff:
I believe an implementation like https://github.com/Moon-0xff/gnome-mpris-label/pull/5/commits/7f134f030c329d37a0c69249ae0d51cd4c57b625 could work too, but if we fail to guess the icon name, instead of throwing an error, it loads an "error icon", that's not good.

Batwam:
Yeah, I saw the error icon in the original iteration but not so much lately. It's useful for debugging to be honest. Once the rest is implemented, we should do a final check and try to intercept it and return a blank instead but as long as we pick an icon from the running apps, I suspect that the risk of running into this will be minimal.

I'm working my way through the issues as the extension is pretty broken right now. I'll push small fixes mostly for error handling to avoid pushing large chunks of code for while you are potentially working on the same parts. Not planning to do any major refactoring.

Moon-0xff:
No, I'm not working on the same part for this exact reason. So feel free to push your commits.

Moon-0xff:
It looks like i need to scramble getIcon.

Batwam:
No worries, I won't be touching the code for a while so go for it!

Moon-0xff:
Ok. This will be destructive, but only for the time being

Moon-0xff:
Don't bother with testing it.
So `getIcon` is a three part function
1. Get the variables
2. Try to guess
3. Get the icon

Our initial guessing game looks misguided, maybe not. But i feel like this part:
```
              else if ( (DBusAddressMatches || DBusDesktopEntryMatches) != null ){
               if (DBusAddressMatches != null){
                       bestMatch = DBusAddressMatches[0][0];
               }
               else if (DBusDesktopEntryMatches != null){
                       bestMatch = DBusDesktopEntryMatches[0][0];
               }
       }
```
is unnecessary

and this one:
```
       // If both returned matches, compare them
       else if ( (DBusAddressMatches != null) && (DBusDesktopEntryMatches != null || undefined) ){
               //how?
               bestMatch = DBusDesktopEntryMatches[0][0];
       }
```
should be a function

Moon-0xff:
OK, so why all of this? Because the function i added uses the same variables, gets a match, then returns an icon.

But really all that it should do is return the most likely match. or a weighted list of likely matches. That's how [DesktopAppInfo.search()](https://gjs-docs.gnome.org/gio20~2.66p/gio.desktopappinfo#function-search) works.

And guessingGame should compare both weighted lists and return the most likely match or a better weighted list.

Moon-0xff:
Hopefully this makes some sense (and works)

icons.js
```diff
@@ -87,3 +70,35 @@ function matchWithDesktopEntries(suspectAppName){
...
+	if (addressMatch == null && DBusEntryMatch == null || undefined)
+		return
+
+	//let's assume DBusEntry is better (if exists)
+	if (DBusEntryMatch)
+		bestMatch = DBusEntryMatch[0][0];
+	else
+		bestMatch = addressMatch[0][0];
+
+	//now let's try to find bestMatch in runningApps
```
Moon-0xff: And finally return `bestMatch`

Batwam:
ok, couple of thoughts:
- i tried to fix up the fact that the rhythmbox "fallback" icon wasn't going away when a player was started. I think that this is fixed now.
- firefox (snap) and spotify (flatak) are ok. google-chrome (deb) isn't. it worked with the original implementatation though. I tried to do a manual swap of `suspectAppName` from chromium to `google-chrome` but Gio.DesktopAppInfo.search() isn't finding it... perhaps Gio.DesktopAppInfo.search() won't be able to find it? Is there a way to get the full list of Desktop apps to check what's in the list?
- what do we want to do if the user doesn't have rhythmbox instead. Right now it's returning the error icon.
- Could we also hide the icon when paused? I had this working earlier today but the move of `function removeTextPausedIsActive` to label.js has broken this somehow. I don't really want to build a duplicate function to implement the delay as it seems a bit wasteful.
- is Gio.DesktopAppInfo.search searching through all the apps installed? if so, is there a way to only get apps which are actually launched?

Moon-0xff:
> i tried to fix up the fact that the rhythmbox "fallback" icon wasn't going away when a player was started. I think that this is fixed now.

Now completely fixed (hopefully) https://github.com/Moon-0xff/gnome-mpris-label/pull/5/commits/cc724d971663f283577ae10024afacaab79d7085

> Is there a way to get the full list of Desktop apps to check what's in the list?

[Try with this](https://gjs-docs.gnome.org/gio20~2.66p/gio.appinfo#function-get_all)

> what do we want to do if the user doesn't have rhythmbox instead. Right now it's returning the error icon.

Now the fallback is an icon included with GNOME https://github.com/Moon-0xff/gnome-mpris-label/pull/5/commits/bb9c18bc34c6a5dff0bdbd91233719eab30295a7

> Could we also hide the icon when paused? I had this working earlier today but the move of function removeTextPausedIsActive to label.js has broken this somehow.

Fixed it https://github.com/Moon-0xff/gnome-mpris-label/pull/5/commits/cc724d971663f283577ae10024afacaab79d7085

>  if so, is there a way to only get apps which are actually launched?

a convenient way no, but surely is possible

Batwam:
I made an update to the mechanism as I was getting some issues with Gio.DesktopAppInfo.search returning multiple matches. it still worked because of the `compareMatches` fallback but I wanted to make sure both method return a valid result. Based on this, firefox (snap), rhythmbox (deb) and spotify (flatpak) all work. google-chrome (deb) is returning the firefox icon for some reason but that's a story for another day...

icons.js
```diff
@@ -0,0 +1,89 @@
...
+		//how?
+		bestMatch = DBusDesktopEntryMatches;
+	}
+
+	return bestMatch
+}
+
+function matchWithDesktopEntries(suspectAppName){
+	let matchedEntries = Gio.DesktopAppInfo.search(suspectAppName);
+
```
Batwam, line 83: other browser might also identify as chromium?

icons.js
```diff
@@ -0,0 +1,94 @@
...
+	if(!matchedEntries.length === 0)
+		return matchedEntries[0][0];
+
+	if (suspectAppName == "chromium") //retry with the name google-chrome if as both browsers identify as chromium
+		matchedEntries = Gio.DesktopAppInfo.search("google-chrome");
```
Batwam, line 88: other browsers might identify as chromium?

icons.js
```diff
@@ -23,8 +23,9 @@ var getIcon = function getIcon(playerAddress){
 
 	let firstDot = addressWithoutMPRIS.indexOf(".");
 
-	let suspectAppName =
-		addressWithoutMPRIS.substring(0,firstDot);
+	let suspectAppName = addressWithoutMPRIS
+	if (firstDot > 1)
```
Moon-0xff, line 27: I don't get this

Batwam: on my system at least, spotify and rhythmbox identify as `org.mpris.MediaPlayer2.spotify` and `org.mpris.MediaPlayer2.rhythmbox` respectively. firstDot is equal to -1 in those cases and results in a blank `suspectName`:
image
the if check allows to avoid this:
image

Batwam:
ok, current version works pretty well although i have the feeling that the fallback icon isn't getting recognised on my system. Also, doesn't it look like there is some excessing padding on the right with this icon? it looks much wider than on the left
Screenshot from 2022-11-30 09-15-06

Moon-0xff:
Yes padding is behaving badly, sometimes is applied in the middle, sometimes is applied twice in the left or right.
It looks like `box` isn't a good choice.

Batwam:
yeah, the intent was to hide everything as if the source wasn't active which is also why I updated the code to make sure the icon was also hidden. I was just having second thoughts but I think that it works fine as it is now.

Batwam:
ok, I think that it looks better now: 
image

even with 0 padding, icons seem to have 5pt padding when compared to text which can be completely flush against the side so I have included a mechanism to correct this. I also use `this._onPaddingChanged()` in _init() to avoid writing the same code twice.

have a look and let me know what you think

Batwam:
ok, here is what I'm getting having tried all the mpris sources I could find on my PC:
image

Basically the mpris address is the most reliable source of info. Unfortunately, not all sources provide DesktopEntry and those which do don't necessarily provide is with the full address so we end up with the same level of info as what we get from the address. With chrome where that info would have been useful to work out which browser is actually used, the app doesn't provide DesktopEntry... So if we want to keep things simple, we could drop the DesktopEntry approach altogether and get to the same result. Have you found a case where DesktopEntry was providing a more reliable result than the mpris address info?

Moon-0xff:
> DesktopEntry approach altogether since the info is unreliable and, at best, gives us the same result as the mpris address method

Yes, i agree.
Please test and review this latest commit.
I left two log statements intentionally, you can view them using `$ journalctl --follow | grep mpris-label`

Batwam:
not a big fact of version specific hacks but it looks better now with gnome 3.* specific padding: 
image

icons.js
```diff
@@ -1,10 +1,15 @@
 const {Clutter,Gio,GLib,GObject,Shell,St} = imports.gi;
 
 var getIcon = function getIcon(playerAddress){
-	let icon = new St.Icon({
+	const Config = imports.misc.config;
```
Moon-0xff, line 4: Import statements belong to the top (Line 2)

Moon-0xff:
Honestly i don't like either of the padding fixes. But i can't really complain as they solve the issues.
I believe the current approach to building the layout has some fundamental issues. But that's work for another day.

Batwam:
My understanding is that Gtk is including some default padding ~5-10pts so I think that it's ok for us to manually adjust if we don't want it for our particular application. On the Gnome3 specific stuff, I blame inconsistencies between the GTK versions.

Moon-0xff:
There's probably a way of having the same layout (in both 4x and 3.3x) with less problems.

I'm looking at [Clutter.BoxLayout](https://gjs-docs.gnome.org/clutter11~11_api/clutter.boxlayout) as it has a method to [set the spacing](https://gjs-docs.gnome.org/clutter11~11_api/clutter.boxlayout#method-set_spacing). I haven't tested anything but it looks promising.

Moon-0xff:
Also, the main class inherits from PanelMenu.Button and that probably has it's own rules regarding layout.
I just glanced a bit at [the code](https://gitlab.gnome.org/GNOME/gnome-shell/blob/main/js/ui/panelMenu.js#L95) and it looks like it has some rules regarding layout but probably they don't matter to us.
It also looks like it handles a menu so maybe implementing to-do number 3 will not be hard.

Moon-0xff:
Yes, I'm pretty sure a rewrite will be necessary soon.
Regarding the scope of the extension I'm more inclined to do a featureful extension rather than a minimalistic one.

Moon-0xff:
Can you tell me which application icons are still failing to load on your end?

Batwam:
Every app I tried (basically the 4 listed in the previous screenshot) works fines. My only concern would be with other browsers (gnome-web, brave, opera,...) as I'm not sure how they identify. Have you tried any others?

Moon-0xff:
Yes, VLC and chromium. the VLC icon loads just fine but the chromium icon fails to load and i don't know why.

Moon-0xff:
`Shell.AppSystem.get_default().get_running()` returns a list of running apps.
Then maybe you will want to [get the ids](https://gjs-docs.gnome.org/shell01~0.1_api/shell.app#method-get_id) of each one and compare them with the guess.
I have not tested this, but seems reasonable enough.

Moon-0xff:
[Have you experienced any slow downs, freezes or crashes with the extension enabled?](https://extensions.gnome.org/extension/4928/mpris-label/)

Batwam:
> [Have you experienced any slow downs, freezes or crashes with the extension enabled?](https://extensions.gnome.org/extension/4928/mpris-label/)
no, I haven't

Moon-0xff:
Woah. I installed [the app](https://flathub.org/apps/details/de.haeckerfelix.Shortwave) and just by opening it it slow downs the system to a crawl. And yes, the lag goes away by turning off the extension.

Batwam:
Ok, we need to find a way to make this asynchronous and/or be more careful with the number of dbus calls we make. Does it improve much if you ping every 1000ms instead of 300ms default?

I was actually just looking at something similar on the icon as we are currently resetting the icon on every refresh so I was trying to find a way to keep a record of the previous address/icon and only apply `setIcon()` if they changes.

There is another issue currently where if I launch a second source on top of the first with `autoSwitch`, the label will change but not the icon...

Moon-0xff:
Yes, it needs its own separate thread. And the timing is right, as we are mostly done here. 

> Does it improve much if you ping every 1000ms instead of 300ms default?

Yes it looks like it slows down every time the extension refreshes.

> Ok, we need to find a way to make this asynchronous and/or be more careful with the number of dbus calls we make.

Yes, probably this will help a lot. But my theory for now is the extension waits for a response until a timeout is reached. As the extension is part of gnome-shell, it freezes the entire shell.

Moon-0xff:
[Please add your email to your github account](https://github.com/settings/emails)
I'm going to merge this soon.

Batwam:
Before you merge, have a look. I implemented fixes for the two issues mentioned above:

- only update icon when the address is changing. This uncovered an issue where the player.address wasn't cleared when there was no activePlayers left and it therefore didn't pickup on the fact that the player had changed... long story, see changes around extension.js row 214-218.
- icon wasn't updating when a second source was started on top of the first although the text was updated to the other source. We do occasionally come across inconsistencies as the text and icon updates are handled pretty much independantly. Just need to be careful with edge cases...

Since this is pretty fresh, I'd recommend giving it a couple of days before the merge? In the meantime, increase the default refresh-rate to 1000ms perhaps?

icons.js
Batwam: agreed

Batwam (commented):
for my own understanding, would you mind explaining how [79a0483](https://github.com/Moon-0xff/gnome-mpris-label/pull/5/commits/79a0483cd6a70c5b77abc9636dc15849f7113b7e) fixed you issue? 

I would have thought that since all cases with `!matchedEntries.length === 0` would be covered above and the function exited with `return`, we would only be left with `matchedEntries.length === 0` cases at this point of the code and this addition wouldn't make any difference?

Batwam:
> [Please add your email to your github account](https://github.com/settings/emails) I'm going to merge this soon. (With your approval)

done (I think)

Moon-0xff:
```
if(!matchedEntries.length === 0)
		return matchedEntries[0][0]
```
length of matchedEntry is not 0
```
if (suspectAppName == "chromium")
```
suspectAppName is `chromium`
```
matchedEntries = Gio.DesktopAppInfo.search("google-chrome")
```
therefore it tries to load `google-chrome`
google-chrome is not installed so it doesn't return matches

With the fix it verifies that matchedEntry is equal to 0 first.

Moon-0xff:
> done (I think). Do I also have to untick "Keep my email addresses private" ?
No. it's fine.

Moon-0xff:
> Since this is pretty fresh, I'd recommend giving it a couple of days before the merge? In the meantime, increase the default refresh-rate to 1000ms perhaps?

Don't worry, I'm not going to upload to [ego](https://extensions.gnome.org/). Just merge it to main so we can keep our "development threads" neatly organized. Are you OK with this?

Batwam:
no problem with the merge. 

With regards to the https://github.com/Moon-0xff/gnome-mpris-label/commit/79a0483cd6a70c5b77abc9636dc15849f7113b7e question, I'm a little bit puzzled by the fact that it works but I don't know anything about javascript so that's probably why :-)
It feels a bit odd but since I can reproduce the original issue, you are probably best placed to investigate if you think that there is something odd going on...

Moon-0xff:
Maybe a bug with gjs or something like that. We are done now, so I'm going to proceed with the merge.

PR #9, opened 2022-12-02
failsafe for install.sh

Batwam:
fail-safe against being executed outside the repo directory...

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,6 +16,3 @@ else
 	printf "\n\e[32mAll files copied. \nPlease log out and log back in again to load the extension.\n\n\e[0m"
 fi
 cd $OLDPWD
-
-#keep terminal open if script run from file browser
-$SHELL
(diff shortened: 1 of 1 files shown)


Moon-0xff:
I believe you can "reload" gnome-shell on wayland by quitting the session `$ gnome-session-quit`

Batwam:
yeah, `$ gnome-session-quit` logs you out (technically asks you if you want to log out). I wouldn't include the command from the code as I think that most people probably would probably want to close whatever they have open first.but I could make the text more personalised based on `$XDG_SESSION_TYPE`

Batwam:
Please run it on your end just in case but it's good to go as far as I'm concerned.

Moon-0xff:
I've tested it.
The only *slight* issue i have so far is:
**If you execute it from outside the directory. It cd's you to the repository.**
I have pushed a "fix" for this. Let me know what you think.

Moon-0xff:
Don't bother to write a special case if the script is called from inside. `cd "$(dirname "$0")"` updates `$OLDPWD` regardless.

Moon-0xff:
I've run it from nautilus (Right click -> run as program). The terminal window closes itself.
`$SHELL` also opens a shell inside the current one.
The instructions require a terminal already, so we shouldn't worry too much about this.

Batwam:
That's strange as it worked for me. Yeah, I don't normally run scripts from the file browser but I just thought it might be a but more user friendly.

Moon-0xff:
Great. I'm going to merge this now. Thanks again @Batwam

Issue #6, opened 2022-12-02
Update Installation Instructions


Batwam:
Hello, I would suggest to update the [installation instructions](https://github.com/Moon-0xff/gnome-mpris-label#manual-installation) to let people know about the new install.sh script. Great extension by the way! :-)

Moon-0xff:
I don't feel comfortable guiding people towards the install script since it doesn't have a fail-safe against being executed outside the repo directory.

Also, the main repository is sometimes a few changes ahead of the [ego](https://extensions.gnome.org/) version. We probably want to guide people to install the latest version instead of the latest commit. This will require some changes to the git repository.

Batwam:
yeah, a failsafe like this I guess #9 

I agree that you don't really want to have a development version on the front page. You should probably pull these latest changes into some sort of `testing` branch first.

Moon-0xff:
I prefer the opposite of that. A `stable` or `latest-version` branch.
Having the `stable` branch as the main one will be annoying to us and future contributors.

Moon-0xff:
#9 was merged. latest-version branch was created.
We only need to update the installation instructions now.

Batwam:
I'd suggest something like this:
![image](https://user-images.githubusercontent.com/56710655/205342944-35e42067-5cab-4cc4-a6f4-c74f83fb241d.png)

Moon-0xff:
#10 was merged

PR #10, opened 2022-12-02
Update install instructions

Batwam:
as per #6, proposed update to installation instructions


Moon-0xff:
Let me know what you think.

Batwam:
I wouldn't normally volunteer the testing branch as I assume that those who want it probably know how to get it and those who don't know probably shouldn't use it but I have update the text slightly to make it a bit more fool proof (I think) by having the explainations first.

Moon-0xff:
`main` is the development branch.
`latest-version` is the *stable* branch
I agree `latest-version` is a confusing name.
We should pick a better one. What about `release`? Is that a good name?
`latest-release`?

Batwam:
Ok, that's a bit confusing. I would have thought that the default branch would have been the stable one since it's the one git would clone by default?

I personally think that it would make sense to have `main` for stable branch (the same version which is on `ego`) and `testing` for the development branch one.

Moon-0xff:
Adding `install.sh` retroactively could mess up the commit history. So i will not merge this until v5 is out.

PR #11, opened 2022-12-03
Reduce Mpris requests quantity

Batwam:
attempt to minimise quantity of Mpris requests to minimise system slowdowns with certain offending mpris sources. Note that this is very much work in progress.

TODO: 
- minimised calls to mpris source whenever the data is not in use (e.g. playbackStatus currently called even on non-active source) - ongoing
- split getMetadata(address,field) into getMetadata(address) and getField(field) in order to only request the Metadata from the mpris source once per cycle - could you please help on this? I tried but failed miserably...
- dynamically monitor reponse time and increase refresh-rate accordingly and/or disable AUTO_SWITCH_TO_MOST_RECENT & REMOVE_TEXT_WHEN_PAUSED - pending


Moon-0xff:
Please clean your commit history. The readme changes are already part of #6

Batwam:
ok, let me clean this up first

PR #12, opened 2022-12-04
Speed up extension

Batwam:
Attempt to simplify the code where possible, in particular by minimising DBUS requests. Note, this is work in progress.

### DIFF
--- a/dbus.js
+++ b/dbus.js
@@ -40,15 +40,20 @@ var getDesktopEntry = function getDesktopEntry(playerAddress){
 }
 
 var getDBusList = function getDBusList(){
-	let dBusProxyWrapper = Gio.DBusProxy.makeProxyWrapper(dBusInterface);
+	let dBusProxyWrapper = Gio.DBusProxy.makeProxyWrapper(dBusInterface); 
+	let start_time = new Date().getTime();
 	let dBusProxy = dBusProxyWrapper(Gio.DBus.session,"org.freedesktop.DBus","/org/freedesktop/DBus");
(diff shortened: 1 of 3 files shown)


Batwam:
@Moon-0xff Could you please help with the rewrite of getMetadata to extract the Metadata string only once and then read the varisou fields required from the initial string rather than multiple DBUS requests?

Moon-0xff:
Like this?

Batwam:
Ah ah, I was initially thinking about putting the dictionary into a variable and having a separate function to extract the field value from that dictionary but this is simpler so, yes, that works!

I might be breaking some edge cases along the way but let's see how this goes. I'm focusing on reducing the Mpris calls to the media source primarily as they seem to be the largest bottlenecks in the code. Getting the list of sources isn't slow based on the previous time checks.

Moon-0xff:
I will make some big changes to `main` soon, so clean up this branch and let me know when is ready to merge.

Batwam:
Hey, sorry but I'm out of action with a killer headache thanks to covid and I can barely stand... Would you be able to push your changes to the current version?

The branch is mostly functional although I believe that there is a bit of a regression in the metadata request for shortwave in particular as it won't read the artist properly. I need to see as it could just be an issue with the radio station no broadcasting this info.

As this is still a testing version, could we leave the logs for now? I'd like to see what is being requested and how long it's taking.

Moon-0xff:
Sorry, I didn't quite understand. Do you want me to merge this to the not-yet-created-branch-with-my-changes and leave the logs right?

Moon-0xff:
Well it shouldn't matter too much so i will merge it to not-yet-created-branch-with-my-changes, aka `structure-changes`

PR #16, opened 2022-12-17
update to icon padding mechanism

Batwam:
Hi, minor issue but I noticed that the padding didn't quite work when the icon is on the left. I took the opportunity to rewrite the adjustment to keep things concise as it was basically coding the Max function.

Also, the icon location wasn't loaded properly when opening the Settings so I fixed that too.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -63,15 +63,16 @@ class MprisLabel extends PanelMenu.Button {
 	}
 
 	_onPaddingChanged(){
-		const LEFT_PADDING = this.settings.get_int('left-padding');
+		const ICON_PLACE = this.settings.get_string('show-icon');
+		let LEFT_PADDING = this.settings.get_int('left-padding');
 		let RIGHT_PADDING = this.settings.get_int('right-padding');
 		const SHOW_ICON = this.settings.get_string('show-icon');
 
 		if(SHOW_ICON){
-			if (RIGHT_PADDING < 5)
-				RIGHT_PADDING = 0
+			if (ICON_PLACE == "right")
+				RIGHT_PADDING = Math.max(0,RIGHT_PADDING - 5)
 			else
-				RIGHT_PADDING = RIGHT_PADDING - 5
+				LEFT_PADDING = Math.max(0,LEFT_PADDING - 5)
 		}
 
 		this.box.set_style("padding-left: " + LEFT_PADDING + "px;"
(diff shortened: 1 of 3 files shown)


Moon-0xff:
Hello @Batwam, good work catching these problems but have you tried another approach for this? The icon padding hack works well but it's very fiddly, breaks easily and I'm not very comfortable with it.

Moon-0xff:
I have tried to use [Clutter.BoxLayout](https://gjs-docs.gnome.org/clutter11~11_api/clutter.boxlayout) before with no luck.
Now I'm looking at the [GNOME CSS documentation](https://gjs-docs.gnome.org/css/) and perhaps we can set the style in some other, more reliable way.
The interactive example of [border-spacing](https://gjs-docs.gnome.org/css/border-spacing) looks like it does what we want.

Batwam:
Yeah, it's a funny one as it really looks like the padding is applied inconsistently between icons Vs text and gtk3 Vs 4. It would be interesting to see if other method lead to different results. Having looked at the level of padding in other extensions, a few had similar issues with either excessive and/or inconsistent right/left padding. It just looks like most people don't really care. To be honest, it only really noticeable for people with low padding and a bad case of OCD 😄 

I did also consider setting the min padding to 5 which would eliminate the need for the `Max` formula adjustment but it didn't seem worth reducing choice for people using text only just to remove a fraction of a line of code.

Moon-0xff:
Surely there's a way!

Moon-0xff:
I'm still not convinced with this approach, but it works and i can't come up with a better solution so i will merge it.

Moon-0xff:
I reverted the file mode changes. I don't know why they were committed in the first place... I suppose it was a mistake?

Batwam:
Yeah, I was going to look at that as I didn't recognise these changes. No idea how they go committed as they don't reflect the changes I made.

Batwam:
Ok, the changes to pref.js were due to the gap with the current `main`. What was the problem you found with the other commits?

Moon-0xff:
the file mode bits of `extension.js` and `icons.js` were changed from `644` to `755`
this is the same as executing this command:  `chmod +x extension.js icons.js`

If you don't recall something like this maybe you should check your IDE configurations?

Batwam:
ok, that's pretty odd. I have updated the flags for these two files and pushed the update

Moon-0xff:
Don't worry i reverted the file mode already. Thanks for the PR @Batwam

Batwam:
Ok, no worries

Issue #4, opened 2022-11-24
Todo List - Contributions needed


Batwam:
It's me again!

## Features list for discussion (original list):
- [x] Add an option to include application icons alongside the label (See #5)
- [x] Add a user definable filter list for mpris sources (See #20) - included in v11
- [x] Add menu to select active player by a dropdown menu (See #23)
- [x] Add an option to pause a player by clicking the label (The original item was small in scope because of limitations(776ed060a51db3d8efc81e607f87f8f581a6e183), later widened and added in ce3dc073735db7543970059c2080b0d3cee07592, see #26)
- [x] Make this extension refresh asynchronously (See #7)

My thoughts:
1. Interesting idea, I wonder if there is an automated way to locate/load these icons in gjs. It would be possible to include icon for the usual apps (rhythmbox, spotify, firefox, chromium,...) which would probably easier but I'm not sure if copyrights would allow it and there is a risk the icon shipped may not match the one used in the distro... there may be some content in the [Vitals](https://github.com/corecoding/Vitals) or [runcat](https://github.com/win0err/gnome-runcat) github repo which could be reused to load icons next to the text as they both have that.
2. Could work. Alternatively, a drop down (or equivalent) with common apps might work too.
3. In terms of implementation, there may be some code to be taken from the Vitals extension as it allows to chose content based on a menu activated when clicking the extension: [https://github.com/corecoding/Vitals](https://github.com/corecoding/Vitals)
4. it might get confusing if clicking shows the menu in some cases but pauses in other cases? Including these controls within a menu might be better, something similar to this other discontinued [mpris extension](https://extensions.gnome.org/extension/54/mpris-player-control/)
5. I agree that the current implementation which continuously updates is somewhat suboptimal but I don't know if this is even possibly. If it was, I wonder how much this would really save in terms of resource.

- [x] **Bonus Extra:**
Update extension name/description to make what the extension does more obvious and therefore more popular. I'm not sure how I found it initially but not everyone would be familiar with mpris. A lot of the other mpris extensions are not compatible with Gnome 43 so the market is up for grabs :-) There is a user review along the same lines:
![image](https://user-images.githubusercontent.com/56710655/203795360-55d5ceaa-3f2f-4831-b9ca-ca90afd8989c.png)

## Features list for discussion (later additions):
- [x] Add tooltips in Settings panel to clarify the purpose of the various settings. (See #22) - included in v11
- [ ] Add (regex) text filtering for labels (See #25)
[...]

PR #18, opened 2023-01-23
Filter blacklist Implementation

Batwam:
First pass proposal for filter implementation. This is in relation to https://github.com/Moon-0xff/gnome-mpris-label/issues/4#issue-1463356957

Note that I am only doing a blacklist filtering here. I am not sure why we might require a white list?

The implementation should be self-explanatory but basically, I tried to keep it as simple and robust as possible by simply checking if the `source_name` is somewhere in the blacklist. I tested this and it worked fine. I can't really think of any case which would create "false positives". The added benefit is that if the user takes some freedom on the syntax, adding spaces or using semicolons (or any other separator), it would still work since the `SOURCES_BLACKLIST` syntax which is the one the user enters manually doesn't matter much.

The only small thing I noticed is that if the currently playing source is blacklisted by a user, it won't disappear right away which would be neat. It will have to wait until the source is changed manually/automatically or the extension reloaded. Once it's flushed out, it won't come back. It's a bit of an edge case but I just wanted to point it out in case you notice it too. I had a quick look to see if I could address this quickly but I didn't really want to start updating `extension.js` as this could start introducing some unintended regressions.

Let me know if you want me to update the merge request to be against `main` rather than `filter-list`.

PS: obviously, all chromium based browsers will end up being filtered as the same time but that's kind of inevitable

### DIFF
--- a/players.js
+++ b/players.js
@@ -97,6 +97,16 @@ var Players = class Players {
 
 		this.list = this.list.filter(element => dBusList.includes(element.address));
 
+		let SOURCES_BLACKLIST = this.settings.get_string('mpris-sources-blacklist');
+		this.list = this.list.filter(function(element){
+			let source_name = element.address.replace('org.mpris.MediaPlayer2.','');
+			source_name = source_name.replace(/\.instance.*/g,'');
+			if (SOURCES_BLACKLIST.includes(source_name))
+				return false
+
+			return true
+		});
+
 		let addresses = [];
 		this.list.forEach(element => {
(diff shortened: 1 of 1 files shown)


Moon-0xff:
Thanks @Batwam. Looks and behaves well, i will merge it right now.

I agree a whitelist isn't useful on it's own (i will add something to make it more useful in a bit), and maybe we can make it disappear if we add a call to `this.players.next()` at the end.

PR #21, opened 2023-01-25
Add tooltips to Settings (Rebased) and update ChangeLog

Batwam:
(Updated pull request rebased using latest `main` commits)

Included ability to include tooltips in Settings and started to populate some of them.
image

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -49,13 +49,6 @@ function buildPrefsWidget(){
 //label page:
 	position = 0; //this line this line seems to be unnecessary
 
-	addSubcategoryLabel(labelPage,'Behaviour');
-	addSwitch(labelPage,'auto-switch-to-most-recent','Switch to the most recent source automatically:',"Automatically switch to last playing source");
-	addSwitch(labelPage,'remove-remaster-text','Remove remaster text:',"Remove text \"Remaster\" from metadata fields");
-	addSwitch(labelPage,'remove-text-when-paused','Hide when paused:',"Hide text (and icon) when the source is paused");
-	addSpinButton(labelPage,'remove-text-paused-delay','Hide when paused delay (seconds):',0,10800,"Time lag before hiding the text after the source is paused");
-	addSpinButton(labelPage,'refresh-rate','Refresh rate (milliseconds):',30,3000,"Frequency at which the metadata is updated");
-
 	addSubcategoryLabel(labelPage,'Appearance');
 	addSpinButton(labelPage,'max-string-length','Max string length (each field):',1,150,"Max number of characters to be displayed for each field");
 	addEntry(labelPage,'button-placeholder','Button placeholder (can be left empty):',"Text to be displayed in case of metadata error");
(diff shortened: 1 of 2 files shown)


prefs.js
```diff
@@ -91,11 +86,20 @@ function buildPrefsWidget(){
 	labelPage.attach(visibleFieldsBox,1,position,1,1);
 	position++;
 
-	let showIconComboBox = addStringComboBox(labelPage,'show-icon','Show source icon:',{'off':'','left':'left','right':'right'});
+	let showIconComboBox = addStringComboBox(labelPage,'show-icon','Show source icon:',{'off':'','left':'left','right':'right'},"Show icon next to text");
 	if (shellVersion >= 40)
 		showIconComboBox.margin_end = 36;
```
Batwam, line 91: I am finding that this leads to the ComboBox being misaligned on my system

Batwam:
see proposed updated. I have broken them down as some may be more controversial than others...

Regarding the misaligned drop downs in the Label tab, I have included some comments. With the updates I made, they all align perfectly but your milage might vary...

before:
image

after:
image

prefs.js
```diff
@@ -88,14 +83,24 @@ function buildPrefsWidget(){
 		visibleFieldsBox.append(secondFieldComboBox,true,true,0);
 		visibleFieldsBox.append(lastFieldComboBox,true,true,0);
 	}
+	visibleFieldsBox.margin_start = 30; //include margin on left to align with rest of widgets
```
Batwam, line 86: proposed to include left padding to increase widget width so it drives the window with and ends up aligned on the right

Moon-0xff:
### I don't like the UI design changes submitted.
#### I don't like the reorder of the label section:
I reordered the widgets for aesthetic reasons, i feel they look *nicer and tidier* if the column of widgets (right column) is somewhat ordered by widget type and length.

#### I don't like the filters section ordered in two columns:
Setting the filters require typing, the entry boxes *might* be filled with a lot of text, for that reason the filters section benefits from long entry boxes

Batwam:
That's fine, feel free to pick and choose what works for you, these were mere suggestions, that's why I tried to split into separate commits.

Moon-0xff:
It's my time to use git cherry-pick for the first time then

Batwam:
Either that or `git revert` the changes you don't like so you can still keep the updates as a standalone branch prior to merging?

Moon-0xff:
Done. The cherry pick bypassed the PR though.

For the record:
```
git remote add batwam git@github.com:Batwam/gnome-mpris-label.git
git fetch batwam
git branch prefs-changes
git switch prefs-changes
git cherry-pick batwam/tooltips_test~6
git commit --amend (and changed the commit message)
```

Moon-0xff:
Merged to a new branch instead of main as there's still some changes to do. I will raise a new issue where to discuss this.

Batwam:
I agree with the changes you picked and taking the opportunity to simplify some of the descriptions better explained by tooltip.

The only thing I would stil change is having the Reset button in the Filters tab should be half width for consistency with the other tabs? It's less consistent with the other buttons on the tab but at least it makes the Rest button look the same on all tabs. your call...

PR #19, opened 2023-01-24
Add tooltips to Settings

Batwam:
Included ability to include tooltips in Settings and started to populate some of them.

image

PR #22, opened 2023-01-25
Add tooltips (hints) and other changes to the prefs window

Moon-0xff:
This PR supersedes #21 (and #19 precedes #21)

It adds hints(tooltips) for certain options and other changes to `prefs.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,8 @@ Aditionally:
 
 # Version 10
 - Fixes for GNOME 3.3x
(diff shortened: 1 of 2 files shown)


Moon-0xff:
I think we should use tooltips only when:
* The option isn't self-explanatory (when changing the option values there's no apparent effect)
* There's extra information worth explaining 
* The option explanation is too lengthy to be the description

Moon-0xff:
I think it's ready to merge

PR #23, opened 2023-01-28
Implement Menu for player selection

Batwam:
as per item 4 in Development Roadmap #4 , this is a working version to implement a menu selection for the active player.

Todo List:
- [x] generate menu with list of items
- [x] include link to Settings
- [x] make menu dynamically update based on actual sources
- [x] update menu item formatting to highlight selected/active item Vs other sources (See screenshots below)
- [x] grey out on sources when in Manual mode (instead of italic) to indicate it's info only.
- [x] link grey colour used for source to theme style for non sensitive menu items (or equivalent)
- [x] link item selection event to turn `AUTO_SWITCH_TO_MOST_RECENT` on and off
- [x] link item selection event to actual selection of source (i.e. Clicking Spotify switches the source)
- [ ] ~~move Build_Menu code into separate menu.js file~~
- [ ] check code for edge cases and missing inputs which may make it generate errors/fail (e.g. source disappears while menu is open)
- [x] check compatibility (e.g. on GNOME 3.38, the menu works but the dot doesn't appear)
- [ ] ~~check for impacts/improvements to existing code (e.g. button click to refresh extension position no longer works, new `shortname` item could be used in filter section to remove the address re-formatting lines)~~

known issue:
- [x] Menu is refreshed on each iteration. If an item is highlighted and the mouse left in position, the refresh will be visible as the item will blink as it gets refreshed. This is not visible on other extensions using menus. Need to compare implementation to see how the menu could be updated only on click. Edit: fixed by linking menu generation to mouse click instead of each refresh cycle.

**Auto Selection Mode:** tick for Auto, dot to indicate active source. sources in italic to indicate that they are listed for info only
image
**Manual Selection Mode:** tick mark for selected source
image

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -94,7 +94,6 @@ class MprisLabel extends PanelMenu.Button {
 			new_channels.push(Math.round((channel[0] + channel[1]) / 2));
 		});
 
-		log(new_channels);
 		let mixedColor = Clutter.Color.new(new_channels[0],new_channels[1],new_channels[2],new_channels[3]);
 		let color_str = mixedColor.to_string();
 		this.unfocusColor = color_str.substring(0,7); //ignore alpha channel
--- a/extension.js
+++ b/extension.js
@@ -5,7 +5,6 @@ const {Clutter,Gio,GLib,GObject,St} = imports.gi;
 const Mainloop = imports.mainloop;
 const ExtensionUtils = imports.misc.extensionUtils;
 const CurrentExtension = ExtensionUtils.getCurrentExtension();
-const Lang = imports.lang;
 
 const { Players } = CurrentExtension.imports.players;
 const { buildLabel } = CurrentExtension.imports.label;
@@ -185,7 +184,6 @@ class MprisLabel extends PanelMenu.Button {
 		this.player = this.players.pick();
 		this._setText();
 		this._setIcon();
-		//this._build_menu();
 		this._removeTimeout();
 
 		this._timeout = Mainloop.timeout_add(REFRESH_RATE, this._refresh.bind(this));
--- a/extension.js
+++ b/extension.js
@@ -112,13 +112,13 @@ class MprisLabel extends PanelMenu.Button {
 		this.players.list.forEach(player => {
 			let settingsMenuItem = new PopupMenu.PopupMenuItem(player.shortname);
 
+			if (AUTO_SWITCH_TO_MOST_RECENT){
+				settingsMenuItem.label.set_style('font-style:italic');
+				settingsMenuItem.set_style('color:#7f7f7f')
+			}
+
 			//if item is active player, include DOT if auto mode, CHECK if manual mode
 			if (this.player) {
-				if (AUTO_SWITCH_TO_MOST_RECENT){
-					settingsMenuItem.label.set_style('font-style:italic');
-					settingsMenuItem.set_style('color:#7f7f7f')
-				}
-
 				if (this.player.address ==  player.address) {
 					if (AUTO_SWITCH_TO_MOST_RECENT)
 						settingsMenuItem.setOrnament(PopupMenu.Ornament.DOT)
--- a/extension.js
+++ b/extension.js
@@ -135,11 +135,6 @@ class MprisLabel extends PanelMenu.Button {
 
 				this.players.selected = player; //this.player should sync with this on the next refresh
 				this._refresh();                //so let's refresh right away
-				if (AUTO_SWITCH_TO_MOST_RECENT)
-					this.settings.set_boolean('auto-switch-to-most-recent',false);
-
-				this.players.selected = player;	//this.player should sync with this on the next refresh
-				this._refresh()					//so let's refresh right away
 			});
 
 			this.menu.addMenuItem(settingsMenuItem);
--- a/extension.js
+++ b/extension.js
@@ -99,7 +99,7 @@ class MprisLabel extends PanelMenu.Button {
 		}
 	}
 
-	_buildMenu(){ //https://gjs.guide/extensions/topics/popup-menu.html#popupmenubase
+	_buildMenu(){
 		const REPOSITION_ON_BUTTON_PRESS = this.settings.get_boolean('reposition-on-button-press');
 		const AUTO_SWITCH_TO_MOST_RECENT = this.settings.get_boolean('auto-switch-to-most-recent');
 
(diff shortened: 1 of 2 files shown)


Batwam:
See current version and roadmap above. Feel free to provide any comments/pointer. I have now managed to get the menu to indicate the playing source. Next step is to work on the actions connected to each item. (Edit, I have now also implemented the source switching upon item selection...)

For example, I can read this.settings.get_boolean('auto-switch-to-most-recent') but what's the easiest way to change the value to true/false?

Moon-0xff:
I pushed my changes, i didn't really changed the implementation, just switched how the values are accessed. Inspect the commits for more details. I also removed a lot of the comments, for brevity.

I also connected the auto-switch button to the setting, the auto-switch prevents changing the source manually.

Regarding compatibility, the dot appears now on 3.38, along with the italics and the check.

Batwam:
That's excellent, thanks! I have also updated the todo list at the top with some design decisions we might need to make.

Batwam:
Ah, I was working on this but my github personal token expired! At least it shows we are on the same page. I had some other changes too so I'll try to push that once I've figured out how to update my credentials!

Moon-0xff:
It should only be a matter of [deleting the key from your github account](https://github.com/settings/keys), deleting the expired keys from your .ssh folder, and [generating a new key pair](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)

Batwam:
> It should only be a matter of [deleting the key from your github account](https://github.com/settings/keys), deleting the expired keys from your .ssh folder, and [generating a new key pair](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)

yeah, I had the new token but still needed to re-register it with `git remote set-url origin https://user:key@...`

Batwam:
alright, see commit above. I have added grey colour to the source names when in Auto mode. This still allow selection for direct switch to Manual mode without having to disable Auto_Switch first. (Decided to only change colour to grey to allow direct selection rather than set as `settingsMenuItem.sensitive = false`). I was tempted to make them non-sensitive but selecting a specific source would have required to first untick `Switch Automatically` and then go back to select the source which I felt was a bit cumbersome.
image

Moon-0xff:
It looks good but i wonder if we can get the color from the theme, as if someone has a custom theme installed it might blend with the background.

Also, line 138 to 142 are repeated statements.

Batwam:
Thanks for that, there were some merge conflicts as I merged my changes into yours since they overlapped and it looks like I forgot to clean that part out.

Yeah, there is probably a neater way to call the colour actually recommended by the theme, I'll add this to the todo list.

Moon-0xff:
I had a look on [this page](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/data/theme/gnome-shell-sass/_colors.scss)
And `insensitive_fg_color` might be the one to use. I don't know how to add it to the style though.

Custom styles should have the same name i suppose.

Batwam:
Yeah, custom themes should have consistent naming. If they are missing there is hopefully some sort of built-in fallback mechanism. Now the challenge is to find how to make use of either this colour, or the style it is included in.

Moon-0xff:
I did some detective work on the gnome-shell source code and learned some things. In a nutshell:
* We can get a color declared in the inherited ["theme node file"](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/data/theme/gnome-shell-sass/widgets/) using [lookup_color](https://gjs-docs.gnome.org/st10~1.0/st.themenode#method-lookup_color)
* Our widget doesn't inherit `insensitive_fg_color`

We can, however, get the foreground and background color and mix them in our own, so I did just that.

Batwam:
Nice one, i also did an excessive amount of research on this and did notice yesterday that the colour was actually made of a mix of bg and fg colour `$insensitive_fg_color: mix($fg_color, $bg_color, 50%);` from the colors.scss file https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/data/theme/gnome-shell-sass/_colors.scss

Seems like this is as good as it's going to get!

Batwam:
ok, I had a look at 2 of the outstanding actions and don't think that they need to be addressed:

- moving Build_Menu code into separate menu.js file: it would probably be neater but not strictly necessary and I don't really know how to do it... I'd say no action unless to know how (and want to) tackle it.
- use `shortname` in other parts of the code if it simplified the code: i checked the filter mechanism but the filtering happens very early on as it's based on the `dBusList` which only contains the addresses so these sources get filtered out before they even get a change to get a `shortname` assigned. I'd say no action there.

This basically leaves us with the last step which it to let things settle down for a while and see if we encouter any issues. I haven't had any so far.

Batwam:
quick one. I know that this is out of scope but is there any chance I could slip in some code to remove the "(feat. ...)" and other "Featuring ...)" texts in the label which take too much room and drive me mad? I'd add this as an extra option under Remastered..

I have the code for the filter already and checked against a range of titles I could find on Spotify (**these are not from my personnal library, it was only for testing, I swear!**). It works as intended (doesn't touch the first title, filters out the next 3):
image

Moon-0xff:
Sure, but submit a new PR for that, as this one is almost ready to be merged into `main`

Regarding the leftover actions, i agree with your thoughts.

Batwam:
No worries, I'll park this for now.

Moon-0xff:
I tried to connect the new method to an event, but it looks like there isn't an event for "color scheme changes", I also learned that the foreground and background colors changes depending of where or when in the code we get the theme node.

I would say we are done now, so if you agree i will merge it.

Batwam:
look fine to me. I just had to remove a leftover log statement.

I just pulled the latest change and it's working. Did you re-test on gnome 3.38 considering the recent colour hack?

Moon-0xff:
I develop the extension on GNOME 3.38 so the question should be about 3.36

Moon-0xff:
Well, we are done here, regarding 3.36 let's test it when we are ready for release.

Batwam:
I just fired up the old laptop with Ubuntu 20.04 (Gnome 3.36.8) and everything is working as expected 👍

By the way, the vlc icon shows up too.

Batwam:
One last thing, will you be taking care of the changelog update?

You should probably also update the screenshot (or add a new screenshot) to show the menu and maybe also the settings menu if EGO allows multiple screenshots?

Batwam:
I can give it a go. I'll revert back to the normal Gnome layout with panel at the top for it. What do you want to show? 
- [ ] just the menu opened and closed? 
- [ ] start in Auto mode and select another source to show the icon/title change? I'll load up a few sources sources to show the various players supported.
- [ ] Settings overview as well? I could open the settings from the menu and click each of the 3 tabs
- [ ] Some of the options in action as well (Remaster, icon location, widget location...)? 
- [ ] Full screen grab or just a zoomed in view of the panel area? I gave it a quick try as shown below. The default screencast app resolution seems to be a bit low so it might be best to go fullscreen so it looks decent and gives a general overview of the desktop with the extension?

Screencast from 2023-02-01 10-56-23

Moon-0xff:
I would say it's important that the mouse pointer is shown, also that the resolution is high so the text doesn't look strangely multicolor.

I want to show the extension changing between 3 different sources, showcasing that:
1. The extension supports multiple players
2. You can select between players running simultaneously
3. The icon support
4. Missing/empty fields are simply not displayed

If you leave the controllers of a player in shot you might be able to showcase something else, but the idea is to keep it short.

Moon-0xff:
There's definitely a size limit

PR #24, opened 2023-02-01
Option to filter feat./Featuring text from label (superseded by #25)

Batwam:
As discuss, I'd like to propose an option to remove the "(feat. ...)" and other "Featuring ...)" texts in the label which take too much room and leads to excessively long strings. I've added this as an extra option under the one for Remaster (enabled by default like the Remaster one). I also slightly tweaked the tooltip wording to something which felt a bit clearer and updated the Remaster description for consistency.
image

I checked against a range of titles I could find on Spotify (**these are not from my personnal library, it was only for testing, I swear!** 😅). It works as intended (doesn't touch the first title, filters out the next 3):
image
image


Moon-0xff:
I added a new option to filter "Mix" text too, merged the functions together, and moved the options to the filters page.

I think the layout needs some tweaks, and the wording and "presentation" of the options isn't clear.

All this changes widens the scope of this PR by a lot, so it needs a new title.

Batwam:
Ah ah, if we continue, this extension is going to become larger than Gnome itself, I'm starting to feel bad for the guy who is having to review our updates every week 😅

Good move on combining the filtering functions. It crossed my mind as there is a fair amount of repeated code but I didn't have time for it.

To be honest, there is a lot of junk in song title, I was looking at a playlist and there is a lot of "Recorded at ...", "Live ...". It would almost justify some comma separated list like what we have for the blacklist?

Moon-0xff:
I would *like* to include something like that but not in the way we are doing this.

For once, we could reduce all the code to a single regex (probably), but I'm not skilled enough to do that.  
The best way of including this would be to pass the filtering and regex building to the user, but what are the implications of that or if it's safe to do it i don't know.

Moon-0xff:
I think the ones we have are sufficient. Honestly, if it wasn't for the latest Beatles releases I wouldn't have included the mix one.

Batwam:
Ok, no worries.

I kind of liked the idea of the customisable list as it was fairly consistent with the general "highly customised" approach in genera and the rest of the filters tab in particular. I also felt like I might want to add some more in the future but let's go with the toggles for now if you want.

Moon-0xff:
Ok your idea of including this as a comma separated list started to grow on me, it might need a lengthy explanation though.

Batwam:
Thanks! I think that by including default entries like `remaster,mix ,featuring,feat.` it should relatively self explanatory when it comes to the syntax.

Then we have the tooltip to explain what it does but if we are worries people may not realise there are tooltips, we could add a bit of text before (or after) the field.

Moon-0xff:
If I'm not mistaken, badly formed regex can cause denial of service, i don't know how easy is to mess a regex up to that extent but probably we should put a warning. Better yet, we should encourage the users to test their regex in a terminal running gjs instead. Maybe we can add a test button and do the gjs testing inside prefs.

Batwam:
The thing is that with the way the extension currently runs, whatever is it the field would be used straight away in the next iteration so as soon as they enter it, it's tested.

Frankly, we don't really have to promote this as a regex filter. If there is a risk of crash with people entering the wrong characters, we could perhaps strip out some of the most problematic special characters like `\` and `/` which could get the string to be misinterpreted, then put the filtered string back into the input field so the user can see that the characters been removed?

The best would be to give it a go a see how well it's working. Worst case we scrap it but I'm sure that there is a way.

Moon-0xff:
Ok, let's start another branch and try to do it as a list. If there's too much problems with the idea or implementation, then we can fallback to this branch and ultimately merge this PR.

Batwam:
ok, I have created the [`filtering_test`](https://github.com/Batwam/gnome-mpris-label/tree/filtering_test) branch for this which is forked off the `main` branch. Probably easier just to start from scratch.

PR #26, opened 2023-02-08
Implement Mouse Controls (Superseded by #29 and #30)

Batwam:
As suggested in #4, see proposed implementation to control the volume of the active mpris source by scrolling up/down on the label.
image


Batwam:
updated to include an option to control the global volume rather than the active source only:

Global Volume (replicated default OSD):
image

Active Source Volume (include source name):
image

Batwam:
Regarding sources not giving access to Volume, I'm not sure how to use `Object.keys`. I haven't found any source not providing Volume so far so if you have one, please let me know so I can test. What I noticed which is a bit of an oddity is that chrome allows to change the volume via mpris interface but while it does return a value which goes up and down on request, the change doesn't have any impact on the source volume. 

At the same time, the Settings panel does have a slider to Chrome (one per app actually) which does affect the volume level but doesn't match the volume level provided by mpris...
image

Batwam:
ok, I checked a fiew players. Spotify, Celluloid,Shortwave, Rhythmbox are all of with mpris volume control. Firefox and Chrome give a volume which can be changed but it does nothing to the actual sounds.

I also noticed that if I go into a terminal and type `dbus-send --print-reply --dest=org.mpris.MediaPlayer2.firefox.instance11471 /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Set string:org.mpris.MediaPlayer2.Player string:Volume variant:double:1` , I get `org.mpris.MediaPlayer2.Player:Volume setting is not supported` which could be used if we could get a similar info through the extension. Chrome on the other hand responds as it everything was fine.

Batwam:
Ok, new update. I noticed that firefox will initially return a value `null` for `this.proxy.Volume` so I can make use of that to identify that we don't have volume control for that source. For Chrome however, it will return `1` initially which is a little bit tricky to exclude as it's a realistic value. I think that I might exclude a source if the value is 1 && address includes `instance`.

I could probably make it even simpler and exclude any source with `instance` in its address as it's a browser and it doesn't look like browsers work anyway but it might be a bit too broad and exclude sources which might otherwise work... or exclude if the volume is 1 and the address includes `chromium`... not sure... Any thoughts?

Moon-0xff:
I checked and tested the code a bit. Implementation wise looks good, I'm not too familiar with events so i need to check more on that.

Style wise it looks unnecessarily abridged. I would prefer a slightly more verbose style. More inline with the rest of the code.

I suppose this is heavily influenced by the code of other extension. Care to share a link?

Moon-0xff:
I checked and tested the code a bit. Implementation wise looks good, I'm not too familiar with events so i need to check more on that.

Style wise it looks unnecessarily abridged. I would prefer a slightly more verbose style. More inline with the rest of the code.

I suppose this is heavily influenced by the code of other extension. Care to share a link?

Batwam:
yeah, it's a bit of a patchwork at the moment as I tried to mirrorred the volume setting between `Global` and `Source` in the hope to combine them eventually into a single function. For instance, the VolumeMax for mpris sources is always 1 so we don't technically need a variable for this but it allows to use the same code for the following lines. We could probably summarise each section in 4-5 lines if they are going to remain split.

I have used a little bit of:
- [Volume Scroller](https://github.com/trflynn89/gnome-shell-volume-scroller/blob/main/volume_scroller%40trflynn89.pm.me/extension.js) for the mouse scroll detection and OSD
- [Volume Scroll](https://github.com/krizajb/gnome-shell-volume-scroll/blob/master/volume_scroll%40qord.si/extension.js) for the Global volume but erally only a line or two to work out how to control the global volume with `get_default_sink()`.

Batwam:
Ok, I have re-written the section more based on the second extension's syntax. I also implemented the fallback and some sort of semi-smart to try to work out if a source is able to handle volume control or not. If it's source only and the source isn't capable, the OSD will show a message to explain that the source isn't compatible (so people don't think it's the extension's fault).

Batwam:
It's all working pretty nicely as far as I can tell. In practice, I don't know if it's useful to keep "Source Only" mode without fallback since the alternative with fallback does the same thing but also works with browsers. It feel like Global sound would be the most popular which is why I set it as default but I still like the Source with Fallback mode as it allows the easily fix the volume individually for each.

I still don't know how Gnome controls chrome volumes. I can see that multiple sliders appear if there are multiple tabs with sources. These sliders are certainly doing something but it doesn't appears to be using mpris. 

This reminds me of the whole DesktopEntry data in dbus. It looks like it was going to be great, until we realised that it was only partially implemented by the various apps and we had to develop a workaround...

Moon-0xff:
Read the mpris specification (the freedesktop website one). I remember there's methods to know the mpris capabilities maybe there's one for the volume controls.

Gnome might have it's own way of controlling the audio, and if it's part of the shell then we could import it to the extension and use it.

Other way i can think of is by sending commands to the audio interface directly (pulseaudio,pipewire,jack) but that's too much for a workaround.

Batwam:
Yeah, I really don't want to get into the weeds with pulseaudio/pipewire either! I am not sure how it works in Gnome but there are extensions which give you a menu so adjust sound for various sources so the solution must be out there somewhere.

Moon-0xff:
That extension imports [this code](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/status/volume.js) i haven't read it but might be the way gnome controls the audio.

Batwam:
Yeah, this is the same library I'm importing to control the global volume. I'll have a look to see if this is what is being used to control individual sources too. The problem we might have might be if the volume source is identical different from the mpris source. If we are lucky, they are identified with the same `address` considering that there seems to be a slider appearing not just for the browser but for each source.

Batwam:
ok, I have found how to isolate Left/Middle/Right click. it was a bit fiddly as the menu kept opening but I got there in the end. I have also implemented play/pause toggle on left click. Ultimately, we could have a drop down in preferences for people to decide what they want to do with each button :-)

Other buttons are also recognised and additional functions could added, for instance, I have buttons recognised as 8 and 9 ("upper side" and "lower side") on both of my mouses which could used for prev/next.

Moon-0xff:
That's great! I believed the menu implementation would completely deny the possibility of using left/right click for something else!

Regarding `bind` and `connect` I don't know.  
I know one comes from standard javascript and the other from GObject.

Batwam:
ok, merge conflicts all resolved so that's good.

With regards to figuring out in a neater way which sources can/cannot accept volume change, I've come accross functions like CanPlay as part of this latest implementation which is quite handy to check capabilities. However, it doesn't appear to be anything like CanSetVolume which would have been quite handy

I do note however that the spec says that if Volume is changed, "the signal `org.freedesktop.DBus.Properties.PropertiesChanged` is emitted" https://specifications.freedesktop.org/mpris-spec/2.2/Player_Interface.html#Property:Volume
I was thinking that perhaps, if we set the value to itself and the signal isn't emitted, this could be an indication that the source doesn't accept volume changes?

Do you know of a way to pick up this signal? More generally, this sounds like a way to remove the refresh altogether and only update when DBus tells us there has been a change perhaps?

Moon-0xff:
`PropertiesChanged` is emitted when **any** property is changed so that might give lots of false positives/negatives. Is worth a try though.

It **is** the way of making this extension refreshless, plus the one for the DBusList. For a working implementation checkout [gnome-shell's mpris.js](gitlab.gnome.org/GNOME/gnome-shell/js/ui/mpris.js).

I ~~hope~~ think an implementation of this for the extension isn't difficult. We merely haven't tried. That's another thing for the list.

Batwam:
yeah, I was just looking to see if I could set the volume to itseld and pick and the DBus signal. I'm looking at combining the following two:
- PropertiesChanges signal: https://dbus.freedesktop.org/doc/dbus-api-design.html
- connectSignal (scroll to the bottom): https://extensions.gnome.org/review/7884

It's not working (extension works fine but no signal emitted as far as I can tell). I'll push the code if you want to have a look. Also, I have tried but I haven't figured out how to merge the `mprisInterface`/`entryInterface` and `dBusInterface`/`dBusPropertiesInterface`

Moon-0xff:
I think the signal "name" is different. Search in the mpris.js file something like `propert` that might point you to the exact code to use

Batwam:
That's how it's called in [this example](https://dbus.freedesktop.org/doc/dbus-api-design.html) and when I look in D-Feet:
image

but looking at the extension you linked above, I see that he is using `this._playerProxy.connectObject('g-properties-changed', () => this._updateState(), this)` instead

Batwam:
ok, I've got it. next thing would be to know **which** property actually changed which I was hoping to get from the previous implementation as it had `<arg type="a{sv}" name="changed_properties"/>`

Moon-0xff:
DbusUtils just includes a few helper functions as far as i can tell. Don't get too confused with the fancy and professional looking coding style. You might get a good grasp simply by reading the function and variable names and reading the code as it were plain english.

Batwam:
ok, I've got it working now (just outputs a message in the log for now...). Turns out it was a one liner...

Next thing would ideally be to know **which** property actually changed which I was hoping to get from the previous implementation as it had `<arg type="a{sv}" name="changed_properties"/>`

I can also confirm that the signal is emitted if I change the volume on Spotify but not with Firefox or Chrome :-)
Next step is to find a way to monitor this signal when trying to set a volume...

Moon-0xff:
Maybe populating the parentheses (storing the return values in a variable) in the arrow function thing might give some extra information.

Change `() => callback` to `(something,something2) => callback`

I'm not quite sure if that's real javascript though, maybe i'm confusing things.

Batwam:
Some github server error going on, last commit went through but isn't showing up on github...

Batwam:
ok, I had a go at trying to manually implement a way to identify what changed but keeping the old value and comparing to the new for `Volume`, `PlaybackStatus` and `Metadata`.

Batwam:
Ok so, here is what I'm thinking about so this PR doesn't turn into a 3-week rewrite of the extension.

I think that while we are now closer to a refreshless implementation, we could keep that to a future PR. At the end of the day, it's not like it's sluggish and it may not be as straightforward as we thing.

So, my plan for this PR would be to implement a couple of actions like PlayPause we already have, Previous, Next. We can them put these in drop downs in Settings where people decide what each PRIMARY/MIDDLE/SECONDARY button to (I might add buttons 8/9 too if I can find a proper name for them).

That should do and this way, we can always add extra functionalities like opening the app to the drop down for people to select in the future without breaking everything.

Regarding controlling the browsers tab with whatever way gnome handles it, I am not sure how to implement it (yet). What we have is Global/Source with Fallback/Source without Fallback. Would that be enough for a merge?

Moon-0xff:
Yeah, that sounds about right.

I don't have a preference regarding what PR gets merged first, and the stuff about controlling the volume the way gnome does it can be done later if it's too much trouble.

Moon-0xff:
the extension wasn't loading on GNOME 3.38. Turns out `connectObject` is a rather recent addition to the shell. Checkout [this commit](https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/26235bbe5433c48b81bd3917aaf9ef14ff1929b2?page=2) and [this merge request](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1953) for some information

Moon-0xff:
We can match players with running apps [by matching desktop entries](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/mpris.js#L147)

Batwam:
Nice. Would `app.activate()` also be the key for us to be able to launch the app on click? 

By the way, I noticed that with the mpris sources listed in the Gnome notification area, clicking on Firefox will launch a new instance of firefox rather than the tab where the source is running so the Gnome implementation isn't perfect either.

Moon-0xff:
[activate_window](https://gjs-docs.gnome.org/shell01~0.1/shell.app#method-activate_window) looks like it does what we want.

Batwam:
Ok, that makes sense since the app should already be open.

Batwam:
would you know if `Shell.AppSystem.get_default().lookup_app()` looks through running apps only? If so, I would remove the `get_running()` check as it would be redundant.

also found some info on AppSystem capabilities here: https://www.roojs.org/seed/gir-1.2-gtk-3.0/seed/Shell.AppSystem.html

Batwam:
Interesting, I didn't come across that. Shouldn't we simply concatenate by excluding empty `metadata[field]`? The current implementation assumed as all values are available which, you are right, isn't always the case.

Moon-0xff:
We could cache the result and call `buildLabel` when it's needed. This should be done in `getMetadata`, plus all the `GLib.Variant` stuff.

Moon-0xff:
I started a new branch to implement the basic controls (play/pause, prev and next). It should be easier to merge.

Batwam:
> I started a new branch to implement the basic controls (play/pause, prev and next). It should be easier to merge.

would I need to rebase this PR against your new branch then? Will you be making changes to your branch or this one?

Moon-0xff:
They are not *changes*. Simply *merging the simple parts of this PR earlier*.

The changes we have are release worthy and the time is right. [44 is scheduled for release in a month](https://wiki.gnome.org/FortyFour) and i have the gut feeling we will take about the same time to get a `main` worthy of release again.

So for the sake of not annoying the users with repeated updates **and** of delivering them our work quickly:  
I want to release an update soon.

Batwam:
open app doesn't appear to be straightforwards. I've got it working for most sources but some like Firefox and Rhythmbox don't work. I'm not looking through alternative solutions like this: https://github.com/lucaswerkmeister/activate-window-by-title/blob/main/extension.js

Batwam:
ok, I guess this nuts wasn't hard enough to crack. I worked around it with an alternative method using `global.get_window_actors()` which works for all apps. This works better than the Gnome notification menu as it even opens Firefox properly!
image

I have commented out the original code and it works without it. if you want, we could keep it and have the second method as fallback?

Batwam:
Before I add Seek(), would you consider including mouse scroll as an input to seek and we then have a drop down to say "Mouse control" "Volume/Track Position"?

Moon-0xff:
That's great!

I was also thinking of grouping the controls in another tab, but might be just too few options to justify it.  
We can also add back the `players.next()` method as an option.

Moon-0xff:
The basic controls branch is mostly done but the menu opens with any button click regardless of the selected action.

The branch should be easily merge-able with this one, the branch is basically a sort of a cherry pick of this one, plus the settings part.

Batwam:
I just had a look at the branch and noticed that you already implemented a good chunk of the settings part which is great. On that basis, I won't redoing it here.

What I would change is include the player switch like you said but also even "Menu" to the drop down (if people are happy to drop the menu to use the right click for "next" for instance and rely on "Switch Automatically"). Also, I would add actions for backward (8) and forward(9) mouse buttons. That's 5 drops downs so probably worthy of its own tab.

Finally, if we really wanted to, we could implement mouse scroll control to track position using Seek() and have a drop down for mouse control Volume/Track position? I doubt this would be widely used though so perhaps not worth the effort... but if anything, having the drop down would help show the user that the scroll does something (most people don't realise they can adjust the system volume by scrolling the volume icon in Gnome).

Batwam:
Turns out matchedEntries[0][0] isn't necessarily a good match for the .desktop. I have included a check to pick from the results the .desktop which matches one of the open apps.
image

This means that the previous implementation of _activatePlayer() would have worked with the correct .desktop names since Firefox and Rhythmbox are the two players which required selecting matchedEntries[0][1] so that's probably why they couldn't be activated. I've put it back but left the other one as fallback just in case.

Moon-0xff:
Amazing work you are doing here but I'm starting to get no hopes of merging this PR, it's too big and wide in scope. Can we split it in let's say three different ones?
* Volume scroll
* Open the player window on click
* The rest

That would be a lot easier to merge, review and polish (and everything in between)

I would also propose to create a new branch/PR for a proper, refresh-less label update with the `g-properties-changed` signal and try to get it merged first.

Moon-0xff:
In the spirit of making a clearer commit log (and a more useful `git blame` output) i will push to `main` the `basic-controls` branch with the "menu opens with any click" error intact, as I'm thinking of adding a more robust solution for that and other problems/limitations we deal with because of the underlying button implementation.

Moon-0xff:
Pushed to `main` and started working on the refresh-less implementation

Batwam:
yeah, there are quite a few conflicts now for both pull requests. The refresh-less updated are going to change even more of the underlying code. Shouldn't resolve these conflicts and merge filtering/controls first?

Moon-0xff:
A refresh-less label update shouldn't be a big update. I already have a "complete" version but i haven't tested it and it seems to me I'm missing something. I will push the branch so you can view it.

Batwam:
I like the `vfunc_event` trick! I saw it in the other extension but couldn't work out how to use it as I didn't realise it was a built-in function... this is cleaner indeed.

So, what's the next step? Will you be cherry picking commits from this PR or we should rebase and fix conflicts so it can be merged directly?

Moon-0xff:
> You should add this._buildMenu() under case 'open-menu': in extension.js so the menu gets built when the user clicks. It's only built on init right now I believe.

True!

> Anyway, I'll leave you to it as I'm done with these PRs. Let me know if there is anything I can do to help.

No problem, I will split the PR and solve the conflict issues then.
The best help right now is testing the newest `main`

Batwam:
No problem, I have a VM with just about every media player I could get my hands on ready for testing and another one with the dev version of Ubuntu 23.04 to test gnome 44 as soon as it comes out. If they don't put it in 23.04, we can always check in a Fedora VM.

Moon-0xff:
yes, but this update will be short lived. **At most** we will take a month to update it again, so I don't think we should worry about that.

Batwam:
`_activatePlayer()` simplified to work with Amberol and Rhythmbox when the windows are closed. The .desktop identification is now very reliable since [dc17bc0](https://github.com/Moon-0xff/gnome-mpris-label/pull/26/commits/dc17bc00392c8e43fac7af6300d06e5aa2d93d9a) and apps will be opened then the player class is initialised so this should pick up the .desktop.

PR #29, opened 2023-03-08
Activate Source on Click Option

Batwam:
Simple code implementation to open player on click. As long as the .desktop selection is fixed, a simple .activate() will work reliably so it becomes 2 simple lines of code.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -1,7 +1,7 @@
 const Main = imports.ui.main;
 const PanelMenu = imports.ui.panelMenu;
 const PopupMenu = imports.ui.popupMenu;
-const {Clutter,Gio,GLib,GObject,St} = imports.gi;
+const {Shell,Clutter,Gio,GLib,GObject,St} = imports.gi;
 const Mainloop = imports.mainloop;
 const ExtensionUtils = imports.misc.extensionUtils;
 const CurrentExtension = ExtensionUtils.getCurrentExtension();
@@ -151,6 +151,10 @@ class MprisLabel extends PanelMenu.Button {
 				if(this.player)
 					this.player.goPrevious();
 				break;
+			case 'activate-player':
+				if(this.player)
+					this._activatePlayer();
+				break;
 			case 'open-menu':
 				this._buildMenu();
 				this.menu.toggle();
@@ -160,6 +164,11 @@ class MprisLabel extends PanelMenu.Button {
 				this._refresh();
 				break;
 		}
+	}
+	
+	_activatePlayer(){
+		let playerObject = Shell.AppSystem.get_default().lookup_app(this.player.desktopApp);
+		playerObject.activate();
 	}
 
 	_buildMenu(){
--- a/players.js
+++ b/players.js
@@ -193,16 +193,19 @@ class Player {
 		this.icon = this.getIcon(this.desktopApp);
 	}
 	_matchRunningApps(matchedEntries){
-		let activeApps = Shell.AppSystem.get_default().get_running();
-		for(var i = 0; i < matchedEntries.length; i++) {
-			var desktopFile = matchedEntries[i];
-			for(var j = 0; j < desktopFile.length; j++) {
-				var entry = desktopFile[j];
-				var playerObject = Shell.AppSystem.get_default().lookup_app(entry);
+		const activeApps = Shell.AppSystem.get_default().get_running();
+		let match = "";
+		matchedEntries.forEach((n, i) => {
+			n.forEach((entry, j) => {
+				let playerObject = Shell.AppSystem.get_default().lookup_app(entry);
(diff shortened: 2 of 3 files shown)


Batwam:
by the way, what does `entry.launch` do in row 242 of `players.js`? It looks like it would launch the app but it's odd as it's part of the icon function and also, [Gio.DesktopAppInfo](https://lazka.github.io/pgi-docs/Gio-2.0/classes/DesktopAppInfo.html) doesn't have `launch` but only [launch_action](https://lazka.github.io/pgi-docs/Gio-2.0/classes/DesktopAppInfo.html#Gio.DesktopAppInfo.launch_action) which as I have tested runs actions listed in .Desktop files.

Moon-0xff:
That line was introduced in #27, specifically [this commit](https://github.com/Moon-0xff/gnome-mpris-label/pull/27/commits/5d6beda450f320b99bdc0e5bc996f5d68f30630f)

Batwam:
ok, I should have checked that! This was a testing leftover so I have now removed the line (and confirmed it has no impact).

Are you happy with the way this PR works?

Moon-0xff:
[Using var is problematic due to scope reasons](https://gjs.guide/guides/gjs/style-guide.html#var), and I would prefer `Array.forEach` instead of normal of normal for loops.  
Overall it's good.

Batwam:
Thanks for the feedback, I wasn't aware of the `var` thing so that's good to know. I think that I tried the `for each` originally but couldn't quite make it work. Let me look into this again.

Batwam:
ok done. 

It's a little bit fiddly as matchedEntries is an object and not an array so I had to have 2 nested `forEach` and since I can't exit from a forEach, I'm having to put the result into a variable. the original nest for loop idea came from [here](https://stackoverflow.com/questions/10021847/for-loop-in-multidimensional-javascript-array), the nested forEach loop from [here](https://stackoverflow.com/questions/10021847/for-loop-in-multidimensional-javascript-array/74717414#74717414). All in all, the for loop is shorter and could be made even shorter if I replaced desktopFile with matchedEntries[i] everywhere too but they are equivalent I guess.

Batwam:
Last commit is slightly unrelated but I noticed that the current code prevented the open menu when hide label is active (alo affects play-pause which is more problematic) as it would clear the current player (which is a default option).

Moon-0xff:
De-nested `matchedEntries` for the sake of clarity. Arguably a *worse* implementation but i think it's worth it.

Batwam:
I actually just had a brain wave and realised from this that the first entry of the object is just the key so the result we are looking for is always going to be in `matchedEntries[0][x]` so we only need to loop through `matchedEntries[0]`. See simplified code in next commit. Also, as far as I can tell, you can't exit a forEach loop with return, it will always run its course.

On the last point, every result is unique as far as I can tell as I haven't seen any case of muliple matches so I don't think that there will be a first or last.

Moon-0xff:
`DesktopAppInfo.search` [returns a sorted list of matches bundled by score](https://gjs-docs.gnome.org/gio20~2.0/gio.desktopappinfo#function-search). I suppose there's no guarantee that the *correct* result will always get the highest ~~lower~~ score, but all players that the first match isn't the *correct* result is because the first match was a symbolic link to the second, so maybe it's true that we only need to loop through `[0]`? I'm not sure.

`return entry` will indeed work here, using `x => { //code }` doesn't change the scope, but `(x) => { //code }` does.  
Some info [here.](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)

Batwam:
I think that I spoke too quickly. I just checked again with chrome and it does have `matchedEntries[1][0]`, however, it's not a match. Doesn't really make any difference to the end result but cycling through `matchedEntries[i][j]` maybe be needed. On this basis, do you want to revert back to your previous implementation?
image

Moon-0xff:
Yes, i suppose, if you agree?  
By the way, does `return entry` works on your end? It might be a difference between gjs versions.

Batwam:
On the nested return, as far as I can tell, it will keep cycling whether regardless if `(x)` or not:
image

Moon-0xff:
Does `return match` works if the arrow function is declared with `(entry,match) => {`

Batwam:
that leads to the same result on my end

Moon-0xff:
Agreed, but first: is pick behaving the same way after applying [670f248](https://github.com/Moon-0xff/gnome-mpris-label/pull/29/commits/670f24862b0f8df8f5f0b7bdb96a86080852d36b)? The removed code looks *non-trivial*

Batwam:
Yeah, removing these 2 line worked for me.

Moon-0xff:
Well, i haven't tested it myself but it shouldn't have any unintended effect (i hope). I'm going to merge this now.

PR #31, opened 2023-03-18
Regex Label Filtering (clean)

Batwam:
PR #25 final changes applied to latest `main` for clarity. Note sure why there were merge conflicts as it's a direct copy paste of the old code but here you go.


Batwam:
for info, this is unrelated to this PR but I noticed that the current implementation of the Player filter list doesn't works:
image

Batwam:
for info, I am not convinced that the changes from this PR work [3044260](https://github.com/Moon-0xff/gnome-mpris-label/pull/25/commits/3044260554ec0a38f8706a42920294164628ce4d). As far as I can tell, `filtersPage._settings.connect('changed::user-regex-filter', () => {` isn't getting called when the filter is modified. I can't see any warning sign and the value isn't changed to the code will still try to filter and therefore generate errors in the log:
image

Batwam:
Would it be worth providing the ability to specify which sources the metadata filters should apply to, similarly to the Ignore/Allow lists? Leave blank to apply to all sources.

Batwam:
Following up from the comments in #52, I still think that there is value in extending the range of possible filters, at least for the title field.

Would it be possible to implement a simplified version of this, focussing solely on filtering `title` metadata? if we want to keep things simple for the user, the original #25 PR commit https://github.com/Moon-0xff/gnome-mpris-label/pull/25/commits/e024a8db5b0c2c721759afc410ba0df10b775f29 was based on basic `match`.

Moon-0xff:
Weighing the potential problems, needed changes, and maintainability concerns versus the added value to the extension, I don't think adding this is worth it.

I think the best choice would be to return to the more limited #24 implementation, and make the extension easier to understand and edit.

PR #32, opened 2023-03-18
Fix broken player filtering

Batwam:
Player filtering is currently broken. Please see proposed update.


Moon-0xff:
Creating a dbus proxy without a callback provokes shell freezes when the player is slow to respond (like with shortwave).  
This isn't noticeable this time because `_filterList` is called only when an mpris address is added or removed of dbus but the freezes are still there.

To experience continuous short freezes again: Open shortwave and start typing on the ignore list.

Moon-0xff:
It looks like I simply forgot to match `identity` with blacklist/whitelist. This should be solved just by matching with the right thing:
`.filter(element => whitelist.includes(element.identity.toLowerCase().replaceAll(' ','')));`

**But!** doing this will not solve the problem because `entryProxy` is called with a callback and `identity` isn't ready when `_filterList` is called, therefore it errors out with `element.identity is undefined`

Batwam:
Yeah, I kept getting `element.identity` `undefined`. I figured that it was because this.list only contains the mpris list from DBUS but not the players info.

Batwam:
I guess a quick and dirty alternative would be to use `address` like `chromium` for the purpose of filtering... since the active sources are listed at the top, the user would see that there is no `Brave` or `Chrome` in the list.

Moon-0xff:
Too dirty! Although we haven't released a version with the identity changes so I don't discard the option.

I think the best way would be to emit a signal from `_onEntryProxyReady` or adding a pause in the code using `GLib.timeout_add`

Batwam:
Wasn't there a way to make this asynchronous so it wouldn't freeze the shell? How did you manage to fix the Shortwave slowdowns last time?

Moon-0xff:
Creating `entryProxy` with a callback is making the proxy creation asynchronously, in the sense that doesn't slow down or freeze the main thread while is doing it.  
This however will *delay* the execution of the callback getting us to this scenario, where we need to access a value that's declared in the callback but the callback still hasn't been called, so the value isn't defined.

Moon-0xff:
The execution order is this one:
`_updateList -> _filterList -> _onEntryProxyReady`

But we are expecting values declared in `_onEntryProxyReady` which comes *later* in execution. We are trying to bake the cake before adding any eggs.

Batwam:
Maybe a stupid idea but would there be value in checking the filtering as part of the Player Class after `_onEntryProxyReady`? Perhaps we included a flag `filtered` true/false so when the info becomes available, at the next iteration where the player list is updated, we check this flag (not the identity) and that sources get filtered?

Moon-0xff:
It works right after, but the code needs to be called again, the simplest way is to redefine the ignore list *after* the player is spawned to include it. This also lead me to a crash but let's ignore that fact.

Batwam:
sorry, I messed up as I pushed this PR having modified my main. As the initial proposed solution isn't the effective, I will resubmit using a separate branch and we can work through the proposed solution. See new working branch in #34

PR #34, opened 2023-03-19
Update Broken Player Filtering (new)

Batwam:
Follow-up on #32 with separate rebased working branch. This is a testing branch which doesn't yet solve the issue.

As per other PR, player filtering is no longer effective since it occurs before the `identity` information is made available:
`_updateList -> _filterList -> _onEntryProxyReady`

`_filterList` is only called when:
a. An mpris address is added or removed to the dbus list
b. filter options are modified

This in code translates to:
````
this.dBusProxy.connectSignal('NameOwnerChanged',this._updateList.bind(this)); //and the code in _updateList
this.settings.connect('changed::mpris-sources-blacklist',this._filterList.bind(this));
this.settings.connect('changed::mpris-sources-whitelist',this._filterList.bind(this));
this.settings.connect('changed::use-whitelisted-sources-only',this._filterList.bind(this));
````
Which is part of `_initList`

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -346,8 +346,17 @@ class MprisLabel extends PanelMenu.Button {
 		const REFRESH_RATE = this.settings.get_int('refresh-rate');
 
 		let prevPlayer = this.player;
-		this.players.updateActiveList();
+
+		try {
+			this.players.updateFilterList();
+			this.players.updateActiveList();
+		}
+		catch {
+			; //do nothing
+		}
+
 		this.player = this.players.pick();
+
 		if(this.player != prevPlayer)
 			this._getStream();
 
--- a/players.js
+++ b/players.js
@@ -115,9 +115,6 @@ var Players = class Players {
 		dBusList.forEach(address => this.unfilteredList.push(new Player(address)));
 
 		this.dBusProxy.connectSignal('NameOwnerChanged',this._updateList.bind(this));
-		this.settings.connect('changed::mpris-sources-blacklist',this.updateFilterList.bind(this));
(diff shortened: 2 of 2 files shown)


Batwam:
quick fix would be to update filter on `updateActiveList()`

Batwam:
Ok, I tried to do a simple fix by including the filter update in `updateActiveList()`. It worked for the list but the label no longer showed the song info so I reverted as I had to go do something else and didn't want to leave a broken implementation.

This still could be a solution to explore if we can't work out a neater way to trigger the filter callback.

Batwam:
ok, I updated the code to call `_filterList()` on `updateActiveList()` which means that it runs on every iteration so if it doesn't work the first time around, it will later.

I also noticed that if a source was playing and gets filtered, it stays in the label so I cleared `this.selected` if it's not part of the player list.

Batwam:
By the way, have you considered increasing the default refresh delay from 300ms to 500ms or even 1s? that would reduce the number of cycles quite a bit over time and I doubt anyone would notice.

Moon-0xff:
Calling `_filterList` inside `updateActiveList` is a cheat! But `updateActiveList` is in itself a cheat. I'm not against this approach but we should definitely seek to replace it later (alongside `updateActiveList`) 

If we follow this implementation it would be better to call `_filterList` directly from `_refresh` and drop the underscore, for clarity.

Moon-0xff:
Compared against past implementations, `_refresh` isn't a slow function , and it's still tied to refreshing the label (which we want to refresh reasonably fast), so I don't see the point of increasing the refresh delay.

Issue #33, opened 2023-03-18
Spot Player incorrect icon


Batwam:
I just tried to install [Spot](https://flathub.org/apps/details/dev.alextren.Spot) to see how it would behave. I can't get play music (authentication failed!) so I can't see if it works but noticed that it loaded the Spotify icon: ![image](https://user-images.githubusercontent.com/56710655/226097910-278fd876-a858-4478-b660-1f705ad71906.png)

After closing Spotify, it kept the icon but when I reloaded gnome, it displayed the correct icon:
![image](https://user-images.githubusercontent.com/56710655/226097997-50ffaa3f-20d4-4b19-9100-35fd6150a830.png)

Just putting this out there for future investigations... must be an issue with the fuzzy match. 

unlikely that people would have both open at the same time but something to keep in mind if there is an easy fix.

PR #35, opened 2023-03-19
Prioritise first icon match (Spot Fix)

Batwam:
With regards to isse #33, upon further inspect, due to the inability of the forEach to exit on match, the icon is set based on the last match.

In the case of Spot, if Spotify is open (unlikely by ok...), we get to a situation where `matchedEntries: dev.alextren.Spot.desktop,com.spotify.Client.desktop`. in `_matchRunningApps()`, this the first entry matches a running app, so does the second one which is taken as the correct result.

We could use a `for` loop with a return instead. In this commit, I have instead implemented a `matchFound` flag so when a match is found, no other match is saved. This successfully fixes the issue and on the basis that `matchedEntries` are already ordered but Gnome in order of probability, I believe that it is fair to prioritise the first match. 

I  don' t believe that this introduces any regression but do let me know if you have any comment.

### DIFF
--- a/players.js
+++ b/players.js
@@ -202,9 +202,11 @@ class Player {
 		const activeApps = Shell.AppSystem.get_default().get_running();
 
 		let match = matchedEntries[0];
+		let matchFound = false;
 		matchedEntries.forEach(entry => {
 			let playerObject = Shell.AppSystem.get_default().lookup_app(entry);
-			if (activeApps.includes(playerObject)){
+			if (activeApps.includes(playerObject) && !matchFound){
+				matchFound = true;
 				match = entry
 			}
 		});


Moon-0xff:
> It's a shame we can't run forEach backwards!

We can by chaining it with `reverse()` but i don't think we should do that.

`for` loops are unnecessarily verbose and except in rare occasions javascript grants us a more easier to read alternative.  
Instead of working around the limitations of `forEach` we should use another of those alternatives, this is written in the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach):
> There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool.

> Early termination may be accomplished with looping statements like [for](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for), [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of), and [for...in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in). Array methods like [every()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every), [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some), [find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find), and [findIndex()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) also stops iteration immediately when further iteration is not necessary.

Batwam:
Ok, it looks like `.find()` might be the way to go.

Batwam:
See proposed implementation using `.find()`. 

Note that I have left the `matchedEntries[0]` fallback but I'm not convinced it is necessary (or preferable) to show the icon of an app which isn't running as it's quite likely to be an incorrect icon. I would personnally prefer returning `undefined` and let the extension return the default icon so we know the app wasn't found.

Moon-0xff:
Maybe it's more likely that `_matchRunningApps` fails than `matchedEntries[0]` is a wrong icon. I really don't know but we didn't crosscheck with running apps until recently and since #27 the only issues with icons so far has been #33

Batwam:
yeah, `_matchRunningApps` is the function we worked on recently. The objective was to filx the funciton for the likes of Firefox and Rhythmbox which return multiple results and `matchedEntries[0]` (actually `matchedEntries[0][0]`) isn't the correct one. The `.desktop` is then used to identify the correct icon and player to activate so we need to make sure this is robust.

Turns out we (I) overlooked the unlikely case where different apps are returned in `matchedEntries` and both are open. This PR should solve this and make the function more robust.

If we keep the `matchedEntries[0]`, it will most likely return something but if we get to this point of the code, it means that it corresponds to an app which isn't currently running and therefore unlikely to be the correct one.

Moon-0xff:
You almost convinced me, it's a pretty interesting design choice, but we can also raise a more explicit signal that something doesn't work with the OSD window manager, and we have been lucky getting correct icons from `matchedEntries[0]`

It all comes down as how likely is `_matchRunningApps` to fail vs how likely is `matchedEntries[0]` to deliver the wrong icon, and we don't know the answer of that.

Batwam:
no worries. Regarding the last point, we know from #29 that Firefox and Rhythmobx have 100% chance of returning the wrong .desktop (on my system). I am not aware of any case where `_matchRunningApps` would fail, but I also didn't foresee this Spot situation. 

I guess worst case scenario would be one where both `_matchRunningApps` && `matchedEntries[0]` fail which should be pretty remote as it would be essentially a situation were the Seach returns a completely incorrect app or the `identity` field isn't populated correctly by the source (e.g. some Browser identifies as Chrome) but we haven't seen this happen (yet).

PR #37, opened 2023-03-21
Enable Back/Forward Mouse action

Batwam:
Enable Back/Forward Mouse action, Defaults to previous/next track. Button codes hard coded to 8/9 respsectively.


Moon-0xff:
It looks like this is branched from #31 instead of `main`, and by the way have you tested the `thumb-controls` branch?

Batwam:
yeah, i messed up again... let me fix this...

PR #38, opened 2023-03-21
Implement mouse thumb bindings

Batwam:
Looks like I can create a pull request from your own changes! Changes look ok, however, on my mouse, 9 is forward and 8 back.

also, don't we need `return Clutter.EVENT_STOP;` each each of the `case 8:`/`case9:`

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -136,9 +136,11 @@ class MprisLabel extends PanelMenu.Button {
 				this._activateButton('right-click-action');
 				return Clutter.EVENT_STOP;
 			case 8:
-				this._activateButton('thumb-forward-action')
-			case 9:
 				this._activateButton('thumb-backward-action');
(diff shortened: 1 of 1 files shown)


Batwam:
Yeah, it's pretty straightforward, probably not worth a full blown branch PR for 2 lines of code (hopefully).

It a bit strange that your buttons might be swapped. I'm judging forward/backward based on the action in the web browser.

I also see from [this page](https://wiki.archlinux.org/title/Mouse_buttons)
> the thumb-tip button is 9, and the inner-thumb button is 8)

Moon-0xff:
They aren't. I didn't test the branch! 😁

Moon-0xff:
I will push the changes then, let me know if something is missing

Batwam:
Yep, fine with me 👍

PR #39, opened 2023-03-22
Widget Controls

orionchikby:
Hi.
For myself, I added the ability to control buttons.
I didn’t really understand the specifics of writing extensions, so I probably didn’t write very well.
But how can an idea come in handy.


Moon-0xff:
Hi @orionchikby  
Are you using a touchscreen or another "alternative" for a mouse?

Moon-0xff:
I'm not sure how "justified" is this change when we already implemented the mouse controls, but I know this is a great change for people not using a mouse, and/or without thumb buttons for next/prev.  

I have been thinking of adding just the next/prev buttons at each side of the label like this: `<< (label) >>`, discarding the pause button. What do you think?

orionchikby:
In a normal situation, I need a pause and a forward button. And where they will be located is already better for you to decide. It's just more convenient for me to see all the buttons at once on the right.
I think buttons are optional (not so important) for your extension

Moon-0xff:
They *are* kinda optional, but we already implemented them and widened the scope of the extension, so this to me looks like a sensible change to add (maybe even an obvious change to add)

>It's just more convenient for me to see all the buttons at once on the right
Yeah, that sounds right.

extension.js
```diff
@@ -105,17 +122,27 @@ class MprisLabel extends PanelMenu.Button {
...
 		if(EXTENSION_PLACE == "left"){
 			Main.panel._leftBox.insert_child_at_index(this.container, EXTENSION_INDEX);
+			if(SHOW_CONTROL){
+				Main.panel._leftBox.insert_child_at_index(this.buttonContainer, EXTENSION_INDEX + 1);
+			}
 		}
 		else if(EXTENSION_PLACE == "center"){
 			Main.panel._centerBox.insert_child_at_index(this.container, EXTENSION_INDEX);
+			if(SHOW_CONTROL){
+				Main.panel._leftBox.insert_child_at_index(this.buttonContainer, EXTENSION_INDEX + 1);
```
Batwam, line 138: wrong location should be `Main.panel._centerBox.insert_child_at_index(this.buttonContainer, EXTENSION_INDEX + 1);`

extension.js
```diff
@@ -105,17 +122,27 @@ class MprisLabel extends PanelMenu.Button {
...
+			}
 		}
 		else if(EXTENSION_PLACE == "right"){
 			Main.panel._rightBox.insert_child_at_index(this.container, EXTENSION_INDEX);
+			if(SHOW_CONTROL){
+				Main.panel._leftBox.insert_child_at_index(this.buttonContainer, EXTENSION_INDEX + 1);
```
Batwam, line 144: wrong location should be `Main.panel._rightBox.insert_child_at_index(this.buttonContainer, EXTENSION_INDEX + 1);`

Batwam:
Could we combine the controls in a single container like this? 
image

extension.js
orionchikby: fixed

extension.js
orionchikby: fixed

orionchikby:
the idea of disabling clicking (add none) and moving everything into a `box` is a good.

extension.js
Batwam: great. I think that you might need to mark this as `resolved` yourself.

Batwam:
yep. `none` is now in the last of options as of last night so just update your `main`. The alternative would be to find a way to separate left click on the label and left click on the controls (I haven't found how).

Batwam:
You'll want to use `this.box.add_child` (check the existing code) and disable the left click unless there is a way to distinguish a left click on the label Vs left click on the controls.

You might also want to check that it still works with icon right/left/none just in case.

extension.js
```diff
@@ -436,3 +454,68 @@ class MprisLabel extends PanelMenu.Button {
...
+					this.player.goNext();
+				break;
+			case 'prev-track':
+				if(this.player)
+					this.player.goPrevious();
+				break;
+		}
+	}
+
+	triggerControl(){
```
Moon-0xff, line 504: this name isn't very descriptive but i can't come up with a better one.

Moon-0xff:
Overall I'm pretty satisfied with your changes, most of the changes i pushed now are about *clarity* which is more of an stylistically choice honestly.

extension.js
Moon-0xff: Renamed to `updateControls`

Batwam:
How about using the same existing `this.box` rather than creating a separate `this.container`?

Moon-0xff:
I think the only things left to do is:
* Connect `show-panel-controls` option so there's no need to reload the shell
* Properly hide the controls when the label is also hidden
* Attach the controls to `this.box` if we want to do that

The last two items might be quite challenging

Moon-0xff:
> How about using the same existing this.box rather than creating a separate this.container?
Might be challenging, and a matter of preference

Batwam:
It's a one liner, just replace `container.insert_child_at_index` with `this.box.insert_child` and turn off the hover thingy. Also need to disable left click action though.

I did previously, you can see it in the screenshot shot I posted in my earlier comments. It also solves Dropbox inserting itself but you lose the left click action.

Moon-0xff:
I foresee a bunch of problems that might provoke, but that might be just pessimism. Feel free to share a diff  
(`$ git diff > diff.diff`)

Moon-0xff:
Also dropping the left click action shouldn't be necessary and we should seek a better approach.

CHANGELOG.md
```diff
@@ -24,7 +24,8 @@ Aditionally:
 - Adds filters to allow/disallow specified mpris sources
 - Adds tooltips(hints) for various options
 
-# Upcoming Version
+# Version 13
```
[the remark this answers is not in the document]
Moon-0xff: It was the way I merged the branches, now fixed.

Batwam:
Ok, let's drop the idea then.

prefs.js
```diff
@@ -219,6 +220,7 @@ function buildPrefsWidget(){
 		settings.reset('thumb-forward-action');
 		settings.reset('thumb-backward-action');
 		settings.reset('volume-control-scheme');
+		showControlComboBox.set_active_id(settings.get_string('show-control'));
```
Moon-0xff, line 223: showControlComboBox isn't an object

extension.js
```diff
@@ -421,18 +443,108 @@ class MprisLabel extends PanelMenu.Button {
 		return Clutter.EVENT_PROPAGATE;
 	}
 
-	_disable(){
-		if(this.icon)
-			this.box.remove_child(this.icon);
+	_disable () {
```
Moon-0xff, line 446: Not a fan of this. Semicolons although optional in javascript are always recommended because of parser/interpreter issues (i think). After a `return` adding or not adding a semicolon shouldn't have any effect.

In any case, I'm only somewhat following the style of the [original extension](https://github.com/mheine/gnome-shell-spotify-label/blob/master/extension.js), and I don't enforce strict rules, although consistency is **pretty** *and* [pretty important](https://gjs.guide/extensions/review-guidelines/review-guidelines.html#basics)

prefs.js
```diff
@@ -164,13 +164,12 @@ function buildPrefsWidget(){
...
-	if(shellVersion < 40){
-		filtersPageSubGrid.margin = 0;
-	}
-	else {
-		filtersPageSubGrid.margin_top = 0,
-		filtersPageSubGrid.margin_bottom = 0,
-		filtersPageSubGrid.margin_start = 0,
+	if (shellVersion < 40) {
+		filtersPageSubGrid.margin = 0
+	} else {
```
Moon-0xff, line 169: Another style change that I'm not a fan of, for the sake of consistency.

extension.js
```diff
@@ -363,31 +362,33 @@ class MprisLabel extends PanelMenu.Button {
...
-		catch {
-			; //do nothing
+		catch(err){
+			log("Mpris _refresh: " + err);
 		}
 
 		this.player = this.players.pick();
 
-		if(this.player != prevPlayer)
+		if(this.player != null | undefined && this.player != prevPlayer) {
```
Moon-0xff, line 388: We don't use `{ }` in single line `if` statements. Again, just following the original extension style.

extension.js
```diff
@@ -363,31 +362,33 @@ class MprisLabel extends PanelMenu.Button {
 	}
 
 	_refresh() {
-		const REFRESH_RATE = this.settings.get_int('refresh-rate');
+		const refreshRate = this.settings.get_int('refresh-rate');
```
Moon-0xff, line 370: options are all uppercase, comes from a time when options where global constants. Again, consistency.

extension.js
```diff
@@ -363,31 +362,33 @@ class MprisLabel extends PanelMenu.Button {
...
 
-			if(this.panelControls && this.player != null | undefined)
-				this.panelControls.updateControls(this.player);
+			if (this.panelControls && this.player != null | undefined) {
+				this.panelControls.updateControls(this.player)
+			}
 		}
-		catch {
-			; //do nothing
+		catch(err){
```
Moon-0xff, line 382: This one was missing because some errors we didn't bother to fix 😄

Moon-0xff:
@orionchikby I suggest you to split your changes into more commits, this simplifies the reviewing process, makes the intent of changes more obvious, and takes better advantage of `git` altogether.

orionchikby:
@Moon-0xff 
sorry, I will answer with one comment.
I don't think you should take my request, because it changes the logic of your widget
1 showControlComboBox -  bad merge, my mistake
2 I don't understand the coding style
  I am using the standard, built into phpstorm which complies with https://standardjs.com/rules.html
  Not using {} is a big pain for me. I'm so used to it that it's very hard to give them up. besides, such blocks are more readable, but this is just my opinion.
3 REFRESH_RATE already exist , and i got the error
4 catch - not needed if everything works well. but since he did not log errors, I could not catch them

thanks for your extension

Batwam:
See gjs guide here: https://gjs.guide/guides/gjs/style-guide.html#rule-list

This page is silent on `if` syntax but does provide some inputs on other aspects.

From these [examples](https://gjs.guide/guides/gobject/basics.html#callback-arguments), single line if statements don't have `{}` but they do end with `;` which we haven't been doing. 

This [Gnome guide](https://wiki.gnome.org/Attic/Gjs/StyleGuide) does the same.

I have to say that it's been giving me some headaches, especially when I insert extra line for testing logs and I'm not used to this syntax either but that seems to be the way Gnome devs want us to write.

orionchikby:
It's strange for me.
But apparently they know better.
I don't like to skimp on curly braces.

Moon-0xff:
This branch is too old to merge well into `main`. However in the future, after we finish the refreshless implementation I want to try to add control buttons again, and I will probably use this PR as reference.

So, although this PR will not be merged, it's a great contribution regardless.
Thanks @orionchikby !

PR #40, opened 2023-03-23
Update Controls Preferences Panel

Batwam:
See proposed cosmetic updates to the Preferences panel:
- add `none` option to disable click action
- group backward/forward click with left/middle/right click

image

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -196,14 +196,15 @@ function buildPrefsWidget(){
 		'next player':'next-player','open app':'activate-player','volume mute':'volume-mute','volume up':'volume-up','volume down':'volume-down','none':'none'
 	};
 
-	addSubcategoryLabel(controlsPage,'Mouse bindings');
+	addSubcategoryLabel(controlsPage,'Mouse click bindings');
 	let leftClickComboBox = addStringComboBox(controlsPage,'left-click-action','Left click action:',buttonActions,undefined);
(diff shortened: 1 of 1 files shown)


Batwam:
Yeah, I don't know. It may sound obvious to us know but I think as especially since these features are somewhat hidden, it probably doesn't hurt to make it as clear as possible.

The word "click" is used for the first left/middle/right. If we change to "button" we should change their description too.

For scroll, we could replace "Mouse scroll" with "Scroll wheel" otherwise?

Moon-0xff:
Well, for me it doesn't trigger a sense of inconsistency, "Mouse click bindings" reads weird, but "Left/Right click" doesn't.  
You can change the first 3 items for consistency if you think it would be better, I don't have an opinion on that.

About the scroll name, it comes from the fact that a mouse wheel isn't the only "thing" that can activate the bindings (same thing with the mouse buttons though, but most peripherals don't have more than two/three buttons), but i don't have a problem with that name.

Batwam:
So the headings would be "Mouse bindings" for buttons and "Scroll bindings" for the wheel? I guess removing "Mouse" makes sense as not everyone will have a mouse, could be trackpad but then shouldn't the first category read "Button bindings"?

Moon-0xff:
Closing this and adding the first commit separately.

Issue #36, opened 2023-03-20
Shell crashes when changing settings


Moon-0xff:
`JS object wrapper for GObject 0x560283d7a0d0 (GSettings) is being released while toggle references are still pending.`

Happens when typing fast.

Moon-0xff:
Bad luck! Merging this broke the pref window for GNOME 44. I did also tried to crash the window without luck. @Batwam have you experienced these crashes? Can you test if not? Just type fast at random in the ignore list.

Batwam:
Let me check that now

Moon-0xff:
Using the previous head though! `git checkout HEAD^`

Batwam:
The pre-#41 merge then right? I haven't updated my fork so I don't have the #41 commit yet. I can test then.

Batwam:
I have to say I never experienced this issue. I just tried the type fast in Gnome 44 using pre-#41 from `main` in my fork and still no crash. You might want to revert then.

What issue are you getting in Gnome 44 post-merge?

Moon-0xff:
I can't seem to replicate the bug now! Except when changing the icon option really fast which isn't surprising. Not really sure why though, maybe e563813ecb67e0a096a6b69a3ddccc5dd2828633 fixed the issue? I'm pretty sure I've experienced crashed after e563813ecb67e0a096a6b69a3ddccc5dd2828633, maybe I'm misremembering? In any case ffec27099904b4ed7ce378160fc38c2a9c237330 needs to be reverted.

the `focus-out` signal doesn't exists in GNOME 44 for some reason (at least that's the error).

Batwam:
Yeah, ok. I think that the more "fancy" bindings we put in, the more maintenance we are likely to get in the future so less is probably more if it's not really an issue.

Moon-0xff:
The reason why i raised this issue was because the bug was painfully frequent, but looks like indeed https://github.com/Moon-0xff/gnome-mpris-label/commit/e563813ecb67e0a096a6b69a3ddccc5dd2828633 solved the issue, so that's good.

How's the extension working for you? Have you tested it in that VM with countless players? Thinking of releasing the new version soon.

Batwam:
I haven't checked for a while as that VM stopped booting for some reason but I'll install a few players in gnome-nightly later this weekend and confirm. Pretty sure it will be fine.

Moon-0xff:
[...]

PR #41, opened 2023-03-23
Mitigation for pref window crashes

Moon-0xff:
As mentioned in issue #36 the pref window crashes when changing settings rapidly, this isn't an issue with most preferences except the ones that require typing.

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -193,8 +193,8 @@ function buildPrefsWidget(){
 		settings.reset('mpris-sources-blacklist');
 		settings.reset('mpris-sources-whitelist');
 		settings.reset('use-whitelisted-sources-only');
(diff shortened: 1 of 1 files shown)


Batwam:
So the idea is to only update on 'focus-out-event' when the user clicks away from the field rather than on every keystroke right?

By the way, is there a way to get a list of all the 'connect' events for a given widget?

Batwam:
Yeah, I didn't know it was possible but figured that might have been what it was doing based on the name. If that's the case that's great. In a way it's similar to the issue we discussed the other day regarding Player filtering and Shortwave.

We might need to do the same when implementing filtering although I quite like being able to see the label update as I type as it gives immediate feedback 🙂

For the events list, I was just wondering how you found out about this one as I was trying to find the equivalent of 'clicked' for a label the other day.

Issue #42, opened 2023-03-25
Metadata not loading on Startup


Batwam:
as discussed in #28 , the metadata is not loading with some apps until the player is stopped/restarted or there is a switch to the next song.

This has been noticed with Spotify and VLC. For info, the current `stable` version doesn't have this Spotify startup loading issue.

Steps to reproduce:
- Check that setting is set to Switch Automatically and Spotify closed
- Launch Spotify
- Start Song

Based on investigations so far, Spotify is selected as active playe but the `metadata` info is empty which leads to a blank label
![image](https://user-images.githubusercontent.com/56710655/227712641-d5c224b2-6fe8-45d0-a6d0-77f8877aaeaa.png)

Batwam:
I have put some logs in variou splaces to see what is running and what isn't and I'm almost more confused now. The nubmers correspond to the row/line in the code. I have also included a log to show if `update()` is being called.

We can see that the playbackstatus is getting updated. However, the metadata remains empty. When `update()` runs because of play/pause, it's still not reading the medata info. However, as soon as the song switches, the metadata appears. Any idea? also, the whole time, Spotify is selected as active player so it's not that:
![image](https://user-images.githubusercontent.com/56710655/227719942-3b2fcfcc-604b-4c18-b11b-486c13ead8da.png)

Batwam:
By the way, `metadata` isn't `null`. It contains 11 keys with `xesam:title`,... but their value is empty.

Moon-0xff:
Reminds me of this https://github.com/Moon-0xff/gnome-mpris-label/pull/3#issuecomment-1303906453

Batwam:
Wow, completely forgot, that's a good memory! I tested `stable` again though and it's working fine. I just can't figure out what is done differently to make it fail now...

Batwam:
Are you able to reproduce it on your end or it's only me?

Moon-0xff:
>  is there anything fundamentally different between stable and main when it comes to metadata loading?
2402390de43ee1455eb17a7f79769022f7ee7a3e and 9b2a1fb3faa42dfbfbe881d3b10720f62076a1ed

Batwam:
results so far:
- `git checkout 461965830e1f0dc3b0ecc5ac497bddb67d167a75`: same issue
- `git checkout ee79793ef76578e1e43310aa94e422d491388bda`: same issue

I'm working through the older versions to seen when the issue started.

[...]

PR #45, opened 2023-04-02
Check for container to prevent occasional extension loading failure

Batwam:
as mentioned in [this comment](https://github.com/Moon-0xff/gnome-mpris-label/issues/42#issuecomment-1484000482), I have been getting errors with the extension being unable to loads when testing previous version or order to investigate #42.

Not sure why we never had the issue before but I believe that this check should be included.

also updated the `if ()` to include spaces as per gnome convention.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -106,15 +106,16 @@ class MprisLabel extends PanelMenu.Button {
 		const EXTENSION_PLACE = this.settings.get_string('extension-place');
 		const EXTENSION_INDEX = this.settings.get_int('extension-index');
 
-		this.container.get_parent().remove_child(this.container);
+		if (this.container.get_parent())
+			this.container.get_parent().remove_child(this.container);
 
-		if(EXTENSION_PLACE == "left"){
+		if (EXTENSION_PLACE == "left"){
 			Main.panel._leftBox.insert_child_at_index(this.container, EXTENSION_INDEX);
 		}
-		else if(EXTENSION_PLACE == "center"){
+		else if (EXTENSION_PLACE == "center"){
 			Main.panel._centerBox.insert_child_at_index(this.container, EXTENSION_INDEX);
 		}
-		else if(EXTENSION_PLACE == "right"){
+		else if (EXTENSION_PLACE == "right"){
 			Main.panel._rightBox.insert_child_at_index(this.container, EXTENSION_INDEX);
 		}
 	}


Batwam:
any issue with this PR?

Moon-0xff:
The error itself is kind of strange: `this.container.get_parent() is null`

How the parent container is null? To me it looks like a deeper problem with initialization or disabling of the extension. Proving that theory needs testing and I haven't done it.

It's also a very infrequent bug that it seems to only affect us as developers, so I don't think there's a need of fixing this quickly.

Batwam:
Ok, no worries. The `this.container.get_parent() is null` error came up as I was trying older versions back-to-back to troubleshoot the Spotify issue. The code change fixed it and it seemed logical to check whether something existed before removing it. I thought it might help avoid other running into similar issues when upgrading the extension but it's hard to replicate.

I haven't seen it recently so I'm ok either way.

Moon-0xff:
I don't see why. The branch still merges cleanly and there's no indication that this is a non-issue now.

However, I don't have plans to test this soon. If it bothers you, you can close this PR and submit it as an issue. No problem.

Batwam:
That's alright, I haven't had the issue for a while. I just didn't want to leave this open indefinitely if we aren't going to merge it. 

I'll go ahead and close this. We'll know where to go if the issue comes up again 😅

Moon-0xff:
I haven't been able to trigger this but now I can imagine how it *could* be triggered:
*If `_updateTrayPosition` is executed **before** the widget is added to the panel.*

which is possible because we connect the signals before adding `this` to the panel:
```js
                this.settings.connect('changed::left-padding',this._onPaddingChanged.bind(this));
                this.settings.connect('changed::right-padding',this._onPaddingChanged.bind(this));
                this.settings.connect('changed::extension-index',this._updateTrayPosition.bind(this));
                this.settings.connect('changed::extension-place',this._updateTrayPosition.bind(this));
                this.settings.connect('changed::show-icon',this._setIcon.bind(this));
                this.settings.connect('changed::use-album',this._setIcon.bind(this));
                this.settings.connect('changed::symbolic-source-icon', this._setIcon.bind(this));

                Main.panel.addToStatusArea('Mpris Label',this,EXTENSION_INDEX,EXTENSION_PLACE);
```

I will merge this then.

Issue #44, opened 2023-03-27
tap to bring to front?


vixalien:
I don't know how sp-tray does it, but when you click on the indicator, it jumps to the spotify app and when you click it again, it jumps back to where you were.

Moon-0xff:
Since yesterday's update you can press right-click (default binding) in the label to raise the window, right-clicking again does nothing. So we are already halfway there.

This was a mere oversight, and shouldn't be a difficult change to add (i hope).  
Answering the question of *how sp-tray does it* i think all comes down to this function:
```js
        // many thanks to mheine's implementation
        showSpotify() {
            if (this._spotiWin && this._spotiWin.has_focus()) {
                // spotify is up and focused
                if (this._notSpotify) {
                    // hide spotify and pull up the last active window if possible
                    Main.activateWindow(this._notSpotify);
                }
            } else {
                // spotify is unfocused or the tray icon has never been clicked before
                this._spotiWin = this._notSpotify = null;
                let wins = global.get_window_actors(); // get all open windows
                for (let win of wins) {
                    if (typeof win.get_meta_window === "function") {
                        if (win.get_meta_window().get_wm_class() === "Spotify") {
                            this._spotiWin = win.get_meta_window(); // mark the spotify window
                        } else if (win.get_meta_window().has_focus()) {
                            this._notSpotify = win.get_meta_window(); // mark the window that was active when the button was pressed
                        }

                        if (this._spotiWin && this._notSpotify) {
                            break;
                        }
                    }
                }
                Main.activateWindow(this._spotiWin); // pull up the spotify window
            }
        }
```

vixalien:
oh that's good to hear. let me re-login to checkout the new features

Batwam:
@Moon-0xff Note that the current implementation in #29 opens the player but does not put the previously focussed window back on with second click.

I had the plan of hiding the window on second click (actually had a TODO note in the code at some point) as it felt like a natural extension of the action until I realised that we don't really know what the user is doing between the first and second click. For waht we know multiple hours could have passed so we could be showing the wrong app. Our implementation is also a little bit different than sp-tray as sp-tray only works with Spotify while mpris-label works with every music player known to man. This means that a user could click once to open Spotify, close it, start Rhythmbox and expect the `Open App` click to open Rhythmbox.

[...]

PR #46, opened 2023-04-02
Restore previously focused window after `Open App`

Batwam:
with reference to #44, this first pass code will refocus the window which was focused before the player is activated.

I haven't quite worked out how to activate alt+tab. For info, the reference code should be here: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/altTab.js

In the meantime, I am saving the previously focused window and re-activating it if the player is focussed.

One thing I would also like to include is minimising the player if it was minimised before being activated.

### DIFF
--- a/players.js
+++ b/players.js
@@ -263,9 +263,8 @@ class Player {
 		if(this.desktopApp){
 			let app = Shell.AppSystem.get_default().lookup_app(this.desktopApp);
 			let focused_window = global.display.get_focus_window();
-			
+
 			if (this.player_window == null || focused_window != this.player_window){//activate player
-				this.previous_app_window = focused_window;
 				if (this.player_window)//only able to record status once window is known
 					this.player_window_minimized = this.player_window.minimized
 
@@ -274,13 +273,13 @@ class Player {
 				return
 			}
 
-			if (this.previous_app_window){//restore previous window
-				if (this.player_window_minimized)
-					this.player_window.minimize();
-				
-				this.previous_app_window.activate(global.get_current_time());
-				// AltTab.AppSwitcherPopup(); //does nothing
-			}
+			if (this.player_window_minimized)
+				this.player_window.minimize();
+
+			// equivalent to Alt+tab. window to be focused is at the end of the list, just before player and Gnome-shell
+			let windows_list = global.get_window_actors();
+			let app_window = windows_list[windows_list.length - 3].get_meta_window(); 
+			app_window.activate(global.get_current_time());
 		}
 	}
 }
--- a/players.js
+++ b/players.js
@@ -168,7 +168,6 @@ class Player {
 
 		const entryWrapper = Gio.DBusProxy.makeProxyWrapper(entryInterface);
 		this.entryProxy = entryWrapper(Gio.DBus.session,this.address, "/org/mpris/MediaPlayer2",this._onEntryProxyReady.bind(this));
-
 	}
 	_onEntryProxyReady(){
 		this.identity = this.entryProxy.Identity;
@@ -195,6 +194,8 @@ class Player {
 			this.desktopApp = this._matchRunningApps(matchedEntries)
 
 		this.icon = this.getIcon(this.desktopApp);
+
+		this.playerWindow = this._guessAppWindow(this.identity);
 	}
 	_matchRunningApps(matchedEntries){
 		const activeApps = Shell.AppSystem.get_default().get_running();
@@ -262,14 +263,13 @@ class Player {
 		if(this.desktopApp){
 			let app = Shell.AppSystem.get_default().lookup_app(this.desktopApp);
 			let focused_window = global.display.get_focus_window();
-			let player_window = this._guessAppWindow(this.identity);
 
-			if (!player_window)
+			if (!this.playerWindow)
 				return
 
-			if (focused_window == player_window){
-				if (this.player_window_minimized)
-					player_window.minimize();
+			if (focused_window == this.playerWindow){
+				if (this.isWindowMinimized)
+					this.playerWindow.minimize();
 
 				//window to be focused is at the end of the list, just before player and Gnome-shell
 				let windows_list = global.get_window_actors();
@@ -277,7 +277,7 @@ class Player {
 				app_window.activate(global.get_current_time()); //equivalent to Alt+tab
 			}
 			else{
-				this.player_window_minimized = player_window.minimized;
+				this.isWindowMinimized = this.playerWindow.minimized;
 				app.activate();
 			}
 		}
--- a/players.js
+++ b/players.js
@@ -1,6 +1,7 @@
-const {Clutter,Gio,GLib,GObject,Shell,St} = imports.gi;
+const {Clutter,Gio,GLib,GObject,Shell,St,Meta} = imports.gi;
 const ExtensionUtils = imports.misc.extensionUtils;
 const CurrentExtension = ExtensionUtils.getCurrentExtension();
+const altTab = imports.ui.altTab;
 
 const mprisInterface = `
 <node>
@@ -269,10 +270,10 @@ class Player {
 			if (this.player_window_minimized)
 				player_window.minimize();
 
-			//window to be focused is at the end of the list, just before player and Gnome-shell
 			let windows_list = global.get_window_actors();
-			let app_window = windows_list[windows_list.length - 3].get_meta_window();
-			app_window.activate(global.get_current_time()); //equivalent to Alt+tab
+			windows_list = windows_list.filter(element => element.get_meta_window().get_window_type() == 0); //filter out non-normal windows
+			let app_window = windows_list[windows_list.length - 2].get_meta_window(); //window to be focused is at the end of the list, just before player
+			app_window.activate(global.get_current_time());
 		}
 		else{
 			if(this.desktopApp){
--- a/players.js
+++ b/players.js
@@ -279,7 +279,15 @@ class Player {
 	}
 	_guessAppWindow(identity){//secondary best-guess method to match player with window
 		identity = identity.toLowerCase();
-		for (const actor of global.get_window_actors()) {
+		
+		for (const actor of global.get_window_actors()) {//first pass exact match
+			const window = actor.get_meta_window();
+			let wm_class = window.get_wm_class().toLowerCase();
+			if (wm_class == identity)
+				return window;
+		}
+		
+		for (const actor of global.get_window_actors()) {//fallback approximate match
 			const window = actor.get_meta_window();
 			let wm_class = window.get_wm_class().toLowerCase();
 			if (wm_class.includes(identity) | identity.includes(wm_class))
--- a/players.js
+++ b/players.js
@@ -259,24 +259,24 @@ class Player {
 			this.proxy.PreviousRemote()
 	}
 	activatePlayer(){
-		if(this.desktopApp){
-			let app = Shell.AppSystem.get_default().lookup_app(this.desktopApp);
-			let focused_window = global.display.get_focus_window();
-			let player_window = this._guessAppWindow(this.identity);
+		let focused_window = global.display.get_focus_window();
+		let player_window = this._guessAppWindow(this.identity);
 
-			if (!player_window)
-				return
+		if (!player_window)
+			return
 
-			if (focused_window == player_window){
-				if (this.player_window_minimized)
-					player_window.minimize();
+		if (focused_window == player_window){
+			if (this.player_window_minimized)
+				player_window.minimize();
 
-				//window to be focused is at the end of the list, just before player and Gnome-shell
-				let windows_list = global.get_window_actors();
-				let app_window = windows_list[windows_list.length - 3].get_meta_window();
-				app_window.activate(global.get_current_time()); //equivalent to Alt+tab
-			}
-			else{
+			//window to be focused is at the end of the list, just before player and Gnome-shell
+			let windows_list = global.get_window_actors();
+			let app_window = windows_list[windows_list.length - 3].get_meta_window();
+			app_window.activate(global.get_current_time()); //equivalent to Alt+tab
+		}
+		else{
+			if(this.desktopApp){
+				let app = Shell.AppSystem.get_default().lookup_app(this.desktopApp);
 				this.player_window_minimized = player_window.minimized;
 				app.activate();
 			}
--- a/players.js
+++ b/players.js
@@ -266,14 +266,18 @@ class Player {
 			
 			if (this.player_window == null || focused_window != this.player_window){//activate player
 				this.previous_app_window = focused_window;
+				if (this.player_window)//only able to record status once window is known
+					this.player_window_minimized = this.player_window.minimized
+
 				app.activate();
 				this.player_window = global.display.get_focus_window();
 				return
 			}
 
-			//if player was minimised before activatePlayer, re-minimise it
-			
 			if (this.previous_app_window){//restore previous window
+				if (this.player_window_minimized)
+					this.player_window.minimize();
+				
 				this.previous_app_window.activate(global.get_current_time());
 				// AltTab.AppSwitcherPopup(); //does nothing
 			}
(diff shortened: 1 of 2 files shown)


Batwam:
some example of use of AltTab here if this makes sense to you: https://github.com/G-dH/advanced-alttab-window-switcher/blob/main/extension.js

Batwam:
ok, we can get an get a list of windows with `global.get_window_actors()`.
Screenshot from 2023-04-02 20-48-44

interestingly enough, this list ends with the player and then Gnome-Shell. Just before the player, 3rd from the end is the window we want to re-focus. See proposed code in next commit. This works around the issue mentioned above and is essentially a manual implementation of alt+tab which avoids having to save the name of the previously focused window.

Batwam:
code confirmed on both x11 and wayland as well as will multiple mpris sources and auto_switch.

Batwam:
With regards to identifying the minimise status, we have previous code which allowed us to identify the window corresponding to a given app here: 7ff41998ea3ebb4875f12987284300c78c93c147

I don't think that it's bullet proof as we had to work with inconsistencies between `identity` and `wm_class` (Mozilla Firefox has wm_class firefox while Chrome has a wm_class google-chrome) which is why we didn't end up using it to activate the player. However, it was close enough as a backup so if we are merely using it to do a best guess for whether the source was minimised or not, knowing that as soon as the app is activated, we will have a reliable ID for the window, that should be good enough?

Moon-0xff:
I was wrong about `AppSwitcherPopup` being part of `Main`, the only reference to the object is on `ui/windowManager`

By the way, I *investigate* the gnome-shell code mostly using grep:
```
$ cd gnome-shell/js/
$ grep -ri AppSwitcherPopup

ui/windowManager.js:            constructor = AltTab.AppSwitcherPopup;
ui/altTab.js:/* exported AppSwitcherPopup, GroupCyclerPopup, WindowSwitcherPopup,
ui/altTab.js:var AppSwitcherPopup = GObject.registerClass(
ui/altTab.js:class AppSwitcherPopup extends SwitcherPopup.SwitcherPopup {
```

Moon-0xff:
Code works, but it's a lot more complicated than it needs to be, I also don't see a real advantage in storing the player window, see commit above.

Moon-0xff:
That's all i guess, please review the current head, i might be missing something.

I'm also curious about this line:
`app_window.activate(global.get_current_time())`
Why it needs the current time? Can you share a link to the documentation?

Moon-0xff:
I'm not very satisfied with the changelog writing, if you can explain the change better, please do

Batwam:
altTab info is available here: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/altTab.js

The reason was am storing the window is because `guessWindow()` isn't as reliable as opening the app using the .desktop for the reasons described [here](https://github.com/Moon-0xff/gnome-mpris-label/pull/46#issuecomment-1493371810) and picking the focussed window just after it is activated. The main reason being that some apps like Amberol or Rhythmbox can be running in the background and have no window at all. In this case, guessWindow() will return null. However, for this particular application where we just want to see if the window is open, then it should work ok. Because of the secondary issue with a window for `spot` which could be mstaken for `spotify` using a simple `.include()`, I would as a minimum suggest to include a first pass identification with exact match `identity` == `wm_class` to minimise the risk of false positives. Finally, I am not sure if there are other issues with wm_class information not being accurate enough to allow identification so if we really have to avoid saving the `window`, we might need to check with some other apps to see what `wm_class` they report.

With regards to minimizing, I'd really like to maintain the functionality as I personally minimise windows to declutter my desktop so maximise/minimise on click is what I would expect the `open app` action to do. This can be done in addition from the existing code and if the user doesn't minimize windows, it won't make any difference so I don't think that it takes away from the experience. Can we put this back in?

Moon-0xff:
Ah, that's a good reason to store the player window, and maybe it's better to do it on the constructor function then.

Regarding the minimizing, I don't feel like both actions can coexist, maybe adding a behaviour option would be the best way to do it but honestly i feel like minimizing windows is pointless, and minimizing windows isn't a vanilla feature of gnome since 3.0 (i think) so it looks like the gnome devs think the same.

Batwam:
While I get that it's not part of vanilla gnome, the `minimize` function still exists and can be enabled in dconf or with gnome-tweaks. It is also enabled by defaults in some major distros including Ubuntu.

If you look at the way I implemented it originally, it simply minimises if the player was originally minimized. For anyone not minimising windows, there is no impact so they can definitely both coexist. The idea is to restore the player window to its original state.

Moon-0xff:
No. I didn't noticed that behaviour. Then sure, sounds nice.

Batwam:
see proposed implementation in the last commit for the minimise if previously minimised.

Note that I haven't touched the `player_window` part so it's still based on `_guessAppWindow()` for now.

Moon-0xff:
Looks good. I don't have anything more to add to this. If you think it's good for a merge, then I will merge it.

Moon-0xff:
Oh! Almost forgot, we should get the player window on init right?

Batwam:
We don't have to.

We could identify the window on init. However, considering some of the lags in loading, I would be somewhat worried that we could face a situation where the user has opened another app by the time the player class is initialised.

Currently, we are scanning for it on every click using `guessAppWindow`. As long as we can trust `guessAppWindow()` we don't need to save it unless we feel like it's inefficient to run it on every click. `app.activate()` relies on the `.desktop` identification which is quite reliable.

Moon-0xff:
the unreliability comes from the fact that some players can play music without an open window, as you mentioned it. So it would be better to get the player window on init when the player window is more likely to be open.

But now that i think of it, if the player window is closed, then the action would be to activate the app, which would (hopefully) bring the window back, so there's not a reason to add this. Furthermore, the new window would probably be different than the one we catch on init, so now the function is broken.

Well then, i would revert this last commit but maybe you have another idea. If you revert the commit and give me the signal i will merge this as previously discussed.

Batwam:
Knowing the window upfront is only needed for to work out if it is minimized or not. This is what I originally introduced guessAppWindow() for and used the term "guess" to highlight that it's not 100% reliable.

Technically we don't need to know the app window upfront to activate it as we use the `.desktop` only for that.

What we could do if we are worried about the `window` changing is overwrite this.player_window using get_focused_window() right after app.activate() just to be safe so the value gets refreshed on app.activate().

Moon-0xff:
Maybe I'm looking at this wrong but it seems to me that we don't need to worry the window changing. In any case is up to you if you want to add more changes to the branch. I definitely feel like this is already complete but there's no harm in adding some safety guards against probable edge cases.

Moon-0xff:
I'm going to bed now by the way, see you in 10 hours.

Batwam:
based on my checks, the rhythmbox window will indeed change if rhythmbox is closed but the player won't be reinitialised so we definitly can't identify the window on `init`. We don't need to worry about the window changing as long as we identify it within `activatePlayer()`. Also, every time we save a value, there is a risk it may not longer be the current one on the next click as we don't monitor (luckily) what the user does between the two click and he/she could have re-spawned the window in the meantime.

On this basis, I think that the current implementation is best until someone runs into a situation where `_guessAppWindow()` fails

Batwam:
just one more thing... if I have the player on a separate workspace, the first click will take me there. The second one however will not take me back to my original desktop. it will focus whatever window (if any) is secondary after the player).

we would ideally need to find a way to list windows from all workspaces, not just the current one? if I press alt+` , I can get a list of all apps so this information should be available somewhere...

Batwam:
arg... I was a bit puzzled by why these conky windows "inserted" themselves so I thought I'd try to open another app on the workspace alongside the player. Turns out that window will before the one before the player so it will switch from the current workspace to the one with the player, then instead of coming back, cycle between the windows on the other workspace.

The only easy way I can think of would be to go back to square 1 and save the focused window. it simplifies the code quite a bit but it's a bit less "smart"... let me know what you think.

Batwam:
note that the Super+Tab seems to be getting the list right. just not sure how to access this list... it must be burried somewhere here: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/altTab.js

Batwam:
ok, I guess I wasn't going to give up that easily...
It looks like `Shell.AppSystem.get_default().get_running()` is giving us the list we need!

players.js
```diff
@@ -1,6 +1,7 @@
-const {Clutter,Gio,GLib,GObject,Shell,St} = imports.gi;
+const {Clutter,Gio,GLib,GObject,Shell,St,Meta} = imports.gi;
```
Moon-0xff, line 1: is `Meta` used?

players.js
```diff
@@ -259,9 +260,43 @@ class Player {
...
+
+		if (!playerWindow)
+			return
+
+		if (focusedWindow == playerWindow){
+			if (this.playerWindowMinimized)
+				playerWindow.minimize();
+
+			let apps = Shell.AppSystem.get_default().get_running();
+			apps[1].activate(global.get_current_time());
```
Moon-0xff, line 274: This approach *almost* works, but it sometimes fails when using more than two workspaces

players.js
```diff
@@ -1,6 +1,7 @@
-const {Clutter,Gio,GLib,GObject,Shell,St} = imports.gi;
+const {Clutter,Gio,GLib,GObject,Shell,St,Meta} = imports.gi;
 const ExtensionUtils = imports.misc.extensionUtils;
 const CurrentExtension = ExtensionUtils.getCurrentExtension();
+const altTab = imports.ui.altTab;
```
Moon-0xff, line 4: we aren't using this

Moon-0xff:
Workspaces didn't even come to my mind! Good catch. As mentioned the approach is somewhat flawed but it looks to be on the right path.

players.js
Batwam: Not anymore, I tested a method which I thought might might need it to define normal windows but that didn't work. It can be removed.

players.js
Batwam: Yeah, same thing, I tried but couldn't figure out how to implement. It can be removed too.

players.js
Batwam: works fine for me. what is the issue?

Batwam: works fine for me. What is the issue?

Moon-0xff: It works as expected, but it isn't what **I** expected.

* Open an application window on two different workspaces
* Open a player in a third workspace
* Click on the window of workspace one
* Change to workspace two **without** interacting with the window
* Activate the 'open app' action (this works as expected)
* Activate the 'open app' action again (takes you to the first workspace)

Now this works as expected, the focus returns to the last focused window, but in this scenario i would expect 'open app' to bring me back to the workspace i was previously on.

players.js
```diff
@@ -259,9 +259,43 @@ class Player {
...
+			const window = actor.get_meta_window();
+			let wm_class = window.get_wm_class().toLowerCase();
+			if (wm_class == identity)
+				return window;
+		}
+		
+		for (const actor of windows_list) {//fallback approximate match
+			const window = actor.get_meta_window();
+			let wm_class = window.get_wm_class().toLowerCase();
+			if (wm_class.includes(identity) | identity.includes(wm_class))
```
Moon-0xff, line 297: the `|` operator is the bitwise OR operator instead of the logical OR operator (`||`)  
I'm not sure if they cause the same effect

Batwam: Good point, I agree that it should be changed

players.js
Batwam: hmmm, maybe worth implementing something like this if the previous app's workspace is different from the last workspace: https://github.com/arjan/gnome-shell-go-to-last-workspace/blob/master/extension.js

but we would need to figure out which workspace the previous app is on. we can do it with the windows but I don't know if it's possible with with the app info. Do you know what info we can extract from the apps listed in `Shell.AppSystem.get_default().get_running()`? if it's only info regarding the app, we might not be able to know on which workspace it it, unless we cross check again the windows list somehow... another rabbit hole!

Batwam: ok, this is pretty niche but we could save the workspace and if it's different from the current, we go there (nothing is focussed so we do nothing except switching), otherwise we open the app as before.

Batwam:
ok, I had to give up on the fancier alt+tab equivalent and simply saving the window as it just makes things much easier and allows not to worry about workspaces, only the window.

The reality is that most people will expect the action to flick between app/workspace in quick succession so this implementation should make sense. I also checked and there is no issue, even if the previously focused window is closed before the button is click again.

Batwam:
more of a nice to have as it's quite robust but I'd also still prefer using `_guessAppWindow()` for initial guess only since we have a reliable way to get the window by checking the focused window right after having activated the player.

players.js
```diff
@@ -259,9 +259,46 @@ class Player {
...
-		if(this.desktopApp){
-			let app = Shell.AppSystem.get_default().lookup_app(this.desktopApp);
-			app.activate();
+		let focusedWindow = global.display.get_focus_window();
+
+		if (!this.playerWindow)
+			this.playerWindow = this._guessAppWindow(this.identity);
+
+		if (!this.playerWindow)
+			return
```
Batwam, line 268: I'm not convinced this is necessary as we can still activate the app based on the .desktop, especially using the latest implementation

Batwam:
Sorry, still doing some checks as the latest changes seem to have had some unintended consequences on some edge cases. Please don't treat this as final.

Moon-0xff:
>  I'm not sure how to do it.

`$ git restore --source d618a4a players.js`

Batwam:
ok, here is a new solution which I believe is more robust. We can positively match between `Shell.AppSystem.get_default().lookup_app(this.desktopApp)` and `global.get_window_actors()` using the `pid` info available in both. I believe that this is more robust than the previous implementation which involved some guesswork I wasn't a fan of. With this method, we have 100% match rate as far as I can tell.

Moon-0xff:
using pids would indeed be more robust. But what happens if two instances of one player(application) are running simultaneously?

I think I know the answer of this one (we can't know which one is the corresponding player unless we are able to get a pid from d-bus) but I'm curious about the output.

Batwam:
yeah, that did cross my mind. This is most likely to occur with browsers and as far as I can tell, both windows have the same PID so they are impossible to tell apart from each other. I'm guessing that a browser window is just like another tab and sharing the same PID.
image

Even if we were to try to save the window info, the reality is that `activate()` wouldn't be able to know what to activate either.

Moon-0xff:
Now, we might be able to get the pid from a player, long ago i did notice this [line](https://github.com/JasonLG1979/gnome-shell-extension-mpris-indicator-button/blob/master/mprisindicatorbutton%40JasonLG1979.github.io/dbus.js#L432) on mpris-indicator

Batwam:
yeah, I think that if we can get the pid, we could potentially do without the `.desktop` altogether, as long as we can extract the icon info we need from the window using `window.get_icon_geometry()`. That could clean up a lot of what we had to do to reliably identify the .desktop with `Gio.DesktopAppInfo.search(this.identity)` and having to match against running apps. I don't believe that the .desktop is used elsewhere.

Probably one for a separate PR through please  :-)
I didn't expect this one to take this long!

Moon-0xff:
Software development! Small changes? think twice!

I think it has some comedic value. A favorite of mine is this one: (08:00) youtube.com/watch?v=T-BoDW1_9P4

Moon-0xff:
I agree to work in a pid-based implementation in another PR, and i have no problems merging this. Although i haven't tested the current version.

Batwam:
ok, please have a play around just to make sure I haven't missed anything. I have checked with the VM containing every mpris player known to mankind but there could still be situations ike the one you picked up previously with the empty workspace.

players.js
Batwam: resolved with latest implementation

Moon-0xff:
Already did, and on my testing behaves as it should. Should i merge this? I **will** merge this.

Batwam:
Ok, let's do it!

PR #50, opened 2023-05-05
Make the scroll up/down actions optional

Moon-0xff:
After looking at @GuilhermeTerriaga fork I noticed the omission of an option to disable the scroll up/down actions.

This PR does just that.

First commit achieves it adding a separate button.  
Second commit removes the button and re-enables the scroll up/down combo boxes.

The combo boxes for up and down are not aligned with the others, I fixed this by changing the halign from END to FILL, but i don't like the bigger combo boxes.

Note: This is WIP, and for now the extension only reacts to the `scroll-up-action` (disables or enables the volume controls)


Batwam:
I can't seem to be able to do `git checkout disable-scroll-actions` but I agree with the logic and the code seems to check out. It's consistent with the comment I made in #51.

Moon-0xff:
> I can't seem to be able to do git checkout disable-scroll-actions

You might need to `git fetch [remote's name]` first, `git fetch --all` as an easier alternative.

PR #56, opened 2023-05-23
Simple User Defined Title Filtering

Batwam:
Third time lucky perhaps with a basic implementation of a user-defined filter for titles under Label in the Preferences:
image

It's basically the same as the current implementation for `remaster` but allows additional strings to be used via coma separated entry like for the whitelist/blacklist. As these strings are eventually turned into a regex expression, I'm also catching invalid filters using the function previously developed.


Moon-0xff:
As mentioned before, exposing the regex to the user introduces a lot of problems, they aren't impossible to solve but certainly isn't worth the effort.  
Any user that wants to add regex filtering or even more complex string manipulation can and is encouraged to edit the source code.

But this is certainly a good addition to the extension, and I think the best way of adding this would be to **not** expose regex to the user.

We might be able to introduce this without much problems if we escape the regex control characters from the user input, although, that would mean that we will need to check if a gjs update adds new ones, shouldn't be a problem!

Batwam:
Ok, what characters would you want to exclude? I guess we could filter those out relatively easily from the `filter` variable. 

We can do this silently in the code before the regex validation check. Alternatively, we could filter and paste the filtered values back into the field to show that they aren't allowed.

Moon-0xff:
The [reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions) probably mentions all control characters.

Regex control codes are mostly ASCII graphical characters, so maybe escaping or forbidding those might be enough.

Batwam:
See proposed implementation. 

I have included a feedback where the filtered list gets put back into the entry so the user can get visual feedback of the fact that the special characters have been removed. This is not striclty necessary as we could do this silently but I felt that it's a bit clearer in case they don't read the tooltip and think that the fact that these characters are ignored is a bug.

Moon-0xff:
Working with a regex made solely of special characters is *very* fun!

As I [previously mentioned](https://github.com/Moon-0xff/gnome-mpris-label/pull/25#issuecomment-1440462776), filtering the values on `prefs.js` would be better so we don't need to pass information from `label.js` back to prefs.

I've tried to do this with the code for #41 with an added workaround for Gtk4 but didn't work.  
Here's a non working [diff](https://gist.github.com/Moon-0xff/61a9100f13845fb9d93d5cee8e86a451).

Here's the regex line (haven't tested it):
```js
const filterRegex = new RegExp('!"\\$%&\\(\\)\\*\\+-\\.\\/:;<=>\\?@\\[\\\\\\]\\^_`{\\|}~','gi');
```
I replaced the regex literal for a `RegExp` call because my text editor was having problems with the syntax highlighting (no wonder).

Worth mentioning that the regex is [ordered](https://upload.wikimedia.org/wikipedia/commons/c/cf/USASCII_code_chart.png).

Batwam:
Thank for the response. I also tried (unsuccessfully) to implement filtering at pref.js level. I based it on the code from [here](https://github.com/Moon-0xff/gnome-mpris-label/commit/3044260554ec0a38f8706a42920294164628ce4d) which is based on `changed`. While I could make it do something like enter the filtered info in a different entry, it wouldn't work if I tried to put the data back into `label-filtered-list`. In addition, I remember you mentioned preferences that using `changed` could lead to a crash. I thought about using `activate` but the problem I have with this method is that, assuming that it only gets triggered when the user presses Enter, it means that the regex would be used if the user doesn't press Enter.

Likewise with your implementation, the regex would only get filtered once the user clicks somewhere else.

With the current implementation on the other hand, the special characters get filtered before they would get used so not risk of them making their way into the code. Also, it will only run as frequently as the extension loops (every 300ms by default) so we should have the issue you mentioned with the repeated keystrokes. Last but not least, it works 😅

Moon-0xff:
The regex for some reason filters `,`, but not `#`. Not sure why.

Batwam:
I haven't spent too much time testing the latest commits but `#` has been removed from the list of flitered character as part of ba169de304eb8d794083883971b939b42aa43da0 so that seems logical?

I personnaly find ```const filteredString = LABEL_FILTERED_LIST.replace(/[`~!@#$%^&*()_|+\-=?;:'".<>\{\}\[\]\\\/]/gi, '');``` more readable than
```
const GraphCharactersRegex = new RegExp('[!"\\$%&\'\\(\\)\\*\\+-\\.\\/:;<=>\\?@\\[\\\\\\]\\^_`{\\|}~]','gi');
const sanitizedInput = LABEL_FILTERED_LIST.replace(GraphCharactersRegex,"")
```
as the first format does avoid the need for escape `\` for a lot of the characters

Batwam:
the above works for me, basically copying the original string back in but keeping the 2 lines as you had it. I am not sure why the previous one removed the `,` characters but the new regex doesn't. 

With regards to `#`, it's debatable whether they should be removed or not as they don't appear to be regex characters anyway? Same thing with `%`, `_` and probably a few others. In fact, based on [this](https://stackoverflow.com/questions/8882624/creating-regexp-with-special-characters), the following might be sufficient: `/([.?*+^$[\]\\(){}|-])/`

Moon-0xff:
I've testing it a bit and the regex behaves well, [as mentioned before](https://github.com/Moon-0xff/gnome-mpris-label/pull/56#issuecomment-1560469956) filtering *just* the control characters shouldn't be a problem.

Overall I think this is ready to be merged, but testing is still needed. It might be better to just push it and daily drive it for a while, it's not like we are pushing it as an update!

Batwam:
Sounds good to me 👍

Moon-0xff:
Done! I think I should have renamed the PR though! It's not very clear what it does.

Issue #7, opened 2022-12-02
System slowdown with some Apps


Batwam:
As reported in a recent comment, some people appear to be experiencing system freezes when the extension is active which go away when turned off:

- e.g.: https://apps.gnome.org/app/de.haeckerfelix.Shortwave/
- I see that some apps seem to struggle with Mpris including Rhythmbox: https://github.com/solus-project/budgie-desktop/issues/794

I wonder if other similar has the same issue. If not, it would be worth checking why not and if they have implemented any workaround. 

It would also be good to know if previous versions of the extension experience the same thing and how much increasing the refresh-rate improve things. 

Ultimately, updating text/icon only upon change would be preferred.

Moon-0xff:
v3 (bffdcd46f4ff7bb286deb8970df80b0e94532570) + Shortwave doesn't freezes the shell. Or at least it isn't too noticeable.

Moon-0xff:
The dBus handling code of [mpris-indicator-button](https://github.com/JasonLG1979/gnome-shell-extension-mpris-indicator-button/blob/master/mprisindicatorbutton%40JasonLG1979.github.io/dbus.js) includes some error checking among other things. We should check if Shortwave works fine with this extension.

Moon-0xff:
Yes, mpris-indicator-button works fine with Shortwave.

Moon-0xff:
[Based on the latest comment on ego](https://extensions.gnome.org/extension/4928/mpris-label/?c=137841) (not mine obviously) it looks like the shell freezes for the amount of time D-Bus gives [until timeout](https://github.com/solus-project/budgie-desktop/issues/794#issuecomment-929718207) with certain applications.

Moon-0xff:
All the applications mentioned are somewhat related to GNOME. can it be a problem with glib, gtk or another GNOME toolkit?

Batwam:
Yeah, it's pretty odd. I just tried Videos/Totem and I can't seem to experience anything. Shortwave is somewhat noticeable with some stutter every time it refreshes.

I think that the best thing to do will be to figure out which specific calls are causing the slowdowns to avoid focusing on the wrong parts.

I have a couple of ideas which might help reduce the number of dbus calls but for initial troubleshooting it would be good to monitor response times over the various dbus calls to understand what is hanging.

Is anyone getting freezes during first source app launch and then it works fine?

Did we say that other similar extensions are fine?

Batwam:
[...]

PR #59, opened 2023-05-29
Add album art

abisammy:
This adds the ability to show the album art in the extension, with customisation settings:
image


abisammy:
Could also be added to main menu of extension?

abisammy:
For a branch with all my changes please see:
https://github.com/abisammy/gnome-mpris-label/tree/abisammy-changes

Batwam:
Hi, I can't comment on whether this will be merged or not but I have a couple of comments/suggestions:
- is this intended to replace the icon or be shown in addition to it? The screenshow above shows both the artwork and Shortwave icon but when I try, the artwork replaces the icon.
- I wonder if there is a "smart" way to set the icon size autimatically based on panel size. With the current implementation, the icon height is limited by the panel size anyway so when that limit is reached it end up increasing the padding.
- might need to check error handling with sources which don't generate artwork like browsers:
image

For info, with these new options, on my system, the Settings window ends up getting a scroll bar on the right as the options no longer fit in the default window size.

abisammy:
Another thing just noticed, when using dash to panel, it doesnt always seem to work with the panel height, therefore icon size is sometimes too small, could be possible to add both percentage (currently implemented) as well as manual override

Batwam:
> Another thing just noticed, when using dash to panel, it doesnt always seem to work with the panel height, therefore icon size is sometimes too small, could be possible to add both percentage (currently implemented) as well as manual override

I noticed this comment from JustPerfection when researching about panel height. Perhaps something to check? https://www.reddit.com/r/gnome/comments/t2hjyg/how_to_get_top_bar_height/hyn24xt

extension.js
```diff
@@ -392,7 +396,18 @@ class MprisLabel extends PanelMenu.Button {
 		if(!ICON_PLACE || !this.player || this.label.get_text() == "" || this.label.get_text() == PLACEHOLDER)
 			return
 
-		this.icon = this.player.icon
+			if(USE_ALBUM && this.player.metadata){
+				const url = stringFromMetadata("mpris:artUrl", this.player.metadata)
+				if(!url) this.icon = this.player.icon
```
Batwam, line 401: see guidance on syntax formatting: https://github.com/Moon-0xff/gnome-mpris-label/pull/39#issuecomment-1490133750

abisammy:
> > Another thing just noticed, when using dash to panel, it doesnt always seem to work with the panel height, therefore icon size is sometimes too small, could be possible to add both percentage (currently implemented) as well as manual override
> 
> Probably best to understand the cause of the issue before introducing a manual hack. I noticed this comment from JustPerfection when researching about panel height. Perhaps something to check? https://www.reddit.com/r/gnome/comments/t2hjyg/how_to_get_top_bar_height/hyn24xt
> 
> Alternatively, how does it look if you don't specify the size and let gnome shell work it out? Do you feel that there is actually a need to specify/reduce the size manually?

Upon further investigation, its dash to panel thats causing the issues. Therefore I think whats best is that set the default to 70 (which works with the gnome default), and add a slider or input for the overide, which can adjust the percentage.

Batwam:
you could also use `this.box.get_height()` with default size 100% which seems to maximise the icon size?

Moon-0xff:
As @Batwam pointed out we already discussed this a while ago, we didn't implemented it at the time but we didn't discard it either.

I will quote myself to save a few keystrokes:
> Sounds real neat specially with shortwave, but we get less than 32px in a 1920x1080 screen, album covers are far too detailed to translate well into such a small amount of pixels.
> 
> As a "general" option i don't think is of too much use, as a "per-application" option sounds good, but we will need to implement "per-application options" first.

I haven't tested this branch yet, but it wouldn't surprised me if even in a 1080p screen the results look good.

As my previous comment says, I think this addition makes more sense as a per-application option, however there's no need to add a whole scheme to handle "per-application options" to do it, a simple entry widget to `prefs` should do the trick.

abisammy:
> Sounds real neat specially with shortwave, but we get less than 32px in a 1920x1080 screen, album covers are far too detailed to translate well into such a small amount of pixels.

I agree with this, may not look good on 1080p monitors

> As my previous comment says, I think this addition makes more sense as a per-application option, however there's no need to add a whole scheme to handle "per-application options" to do it, a simple entry widget to prefs should do the trick.

I agree, would it not be better to have the ability to enable and disable it for all applications (like it is now anyway) though. Additionally, if there is no album cover available, it will default to the application icon like previously.

Moon-0xff:
> would it not be better to have the ability to enable and disable it for all applications (like it is now anyway) though? 

Yes, of course, but also add the ability of enabling it *just* for some players.

> if there is no album cover available, it will default to the application icon like previously

That's a good default. It also depends from the 'show-icon' option, that might not be obvious.  
I suggest to change the label of the option to:  **Use album art *as icon* when available:**

Moon-0xff:
One last thing I have in mind is performance.

Creating and displaying an image is hard work. `_refresh` is called synchronously and is tied to the shell's main loop, this means that a slow function there slows down the entire shell.

If this PR slows down the shell too much, or even creates micro-freezes under certain circumstances we will need to decouple `_setIcon` from `_refresh` first.

abisammy:
1. Cleaned up the settings:
image

2. Will add the ability to filter album art per application (basically copying from filters)

3. So for I haven't noticed a performance decrease, your mileage may vary

abisammy:
Settings now disabled when show source icon is off:
image

Or album art as icon is off:
image

abisammy:
@Moon-0xff Added a whitelist for the apps to display the album art in:
image

Also refacted, removing repeated code!

Batwam:
Could we make this whitelist optional (`false` by default) by having a toggle on the right so all apps display the album by default? 

I have the feeling that people choosing to display the covers would expect it to be shown whenever available so it might be best if they don't have to manually list the apps they want to show it as default.

Batwam:
if we are concerned about updating the icon on every refresh being suboptimal, we could include a check within _refresh() and only update if the label has changed:
```
let prevLabel = this.label.get_text();

[...]

if (this.label.get_text() != prevLabel)
	this._setIcon();
```

extension.js
```diff
@@ -60,6 +60,7 @@ class MprisLabel extends PanelMenu.Button {
 		this.settings.connect('changed::extension-index',this._updateTrayPosition.bind(this));
 		this.settings.connect('changed::extension-place',this._updateTrayPosition.bind(this));
 		this.settings.connect('changed::show-icon',this._setIcon.bind(this));
```
Batwam, line 62:
```suggestion
		this.settings.connect('changed::use-album',this._setIcon.bind(this));
```

extension.js
```diff
@@ -60,6 +60,7 @@ class MprisLabel extends PanelMenu.Button {
...
+		this.settings.connect('changed::use-album',this._setIcon.bind(this));
```
Batwam, line 63:
```suggestion
		this.settings.connect('changed::album-size',this._setIcon.bind(this));
```

Batwam, line 63: missing `this.settings.connect('changed::album-size',this._setIcon.bind(this));`
```suggestion
		this.settings.connect('changed::show-icon',this._setIcon.bind(this));
		this.settings.connect('changed::use-album',this._setIcon.bind(this));
		this.settings.connect('changed::album-size',this._setIcon.bind(this));
```

extension.js
```diff
@@ -60,6 +60,7 @@ class MprisLabel extends PanelMenu.Button {
...
```
Batwam, line 64: missing `this.settings.connect('changed::album-size',this._setIcon.bind(this));`

Batwam:
the changed made to pref.js in [35decaf](https://github.com/Moon-0xff/gnome-mpris-label/pull/59/commits/35decafe4ed0f31225b7f1b8f426790ddeb2af62) are messing up the formatting I'm afraid:
image

abisammy:
Fixed!!
image

abisammy:
> if we are concerned about updating the icon on every refresh being suboptimal, we could include a check within _refresh() and only update if the player or label have changed:
> 
> ```
> let prevLabel = this.label.get_text();
> 
> [...]
> 
> if (this.player != prevPlayer || this.label.get_text() != prevLabel)
> 	this._setIcon();
> ```
> 
> you would also need to revert [52c5907](https://github.com/Moon-0xff/gnome-mpris-label/pull/59/commits/52c59074bd456aa85ef39ec5c1457e203214cdce) which removed the album-size binding

Currently if the filter is empty, all apps are shown by default

Batwam:
yeah, see what you can achieve. 

Honestly, I always felt that we should be able to make this entire extension refreshless considering that we are currently updating the label every 300ms for something when we would probably only really need updating every few minutes (when a song changes) but we never quite got there... might need to be a separate PR though as it might require some extensive testing to make sure we don't break anything in the process...

Batwam:
actually, wouldn't it makes more sense to set the album cover as `this.player.album` as part of `player.js` and trigger is on 'g-properties-changed' event as part of `this.update()` [here](https://github.com/Moon-0xff/gnome-mpris-label/blob/main/players.js#L213)?

Moon-0xff:
Yes, using 'g-properties-changed' in more places other than the `Player` class would be optimal. I've tried to do this but I wasn't able to pass the signal to other objects *correctly*.

In any case the heavy lifting is *most likely* decoupled from the main thread, testing is needed.

There's also the question if we need to tweak some parameters to not waste resources rendering the image, probably not but again, testing is needed.

abisammy:
@Batwam @Moon-0xff I have made a system that is partially event based, still needs to be tested:
https://github.com/Moon-0xff/gnome-mpris-label/pull/60

It uses events when a player is active, however relies on the old refreshing method until a player is open, maybe you know how to make this independent?

abisammy:
What else is needed for this system?

Moon-0xff:
Yes! I don't have any more suggestions for this, I wonder what @Batwam thinks.

Batwam:
feature-wise, I think that the code does what it needs to do. I would however have the following suggestions to make the code a little bit more efficient:

1. We should try to avoid redrawing the icon on every refresh if we can avoid it. The icon doesn't currently resize automatically when the setting is changed. I understand that we will look to make changes to the way the label updates in a separate PR but in the meantime, we should try to implement a "smarter" way to update the icon (and bind `album-size` again if necessary as it was removed in [52c5907](https://github.com/Moon-0xff/gnome-mpris-label/pull/59/commits/52c59074bd456aa85ef39ec5c1457e203214cdce).
2. I think that the album icon could be set in `player.js` and stored under `this.player.album` so we don't have to regenerate it on each refresh. then we can set `this.icon` to `this.player.icon` or `this.player.album` depending on the setting
3. I believe that it would make more sense for the `prefs.js` list to be a blacklist rather than a whitelist for the following reasons:
      - it's more likely that people would want to show the album for most apps rather than only a few
      - default value blank meaning all apps are allowed isn't the normal behaviour for a white list. However, a blank blacklist means all apps are allowed which is the intended behaviour.

feel free to comment if you don't agree, these are only suggestions.

abisammy:
> feature-wise, I think that the code does what it needs to do. I would however have the following suggestions to make the code a little bit more efficient:
> 
> 1. I don't have the number to quantify the impact but it feels like should try to avoid redrawing the icon on every refresh if we can avoid it. I understand that we will look to make changes to the way the label updates in a separate PR but in the meantime, we should try to implement a "smarter" way to update the icon (and bind `album-size` again if necessary as it was removed in [52c5907](https://github.com/Moon-0xff/gnome-mpris-label/pull/59/commits/52c59074bd456aa85ef39ec5c1457e203214cdce).
> 2. I think that the album icon could be set in `player.js` and stored under `this.player.album` so we don't have to regenerate it on each refresh. then we can set `this.icon` to `this.player.icon` or `this.player.album` depending on the setting
> 3. I believe that it would make more sense for the `prefs.js` list to be a blacklist rather than a whitelist for the following reasons:
>    
>    * it's more likely that people would want to show the album for most apps rather than only a few
>    * default value blank meaning all apps are allowed isn't the normal behaviour for a white list. However, a blank blacklist means all apps are allowed which is the intended behaviour
> 
> feel free to comment if you don't agree, these are only suggestions.

Done!
The icon is now set in the player update method. Therefore it will only change when the 'g-properties-changed' event is fired. Additionally the album art setting is now a blacklist!

Batwam:
It could be a resolution thing but I still have an issue with the the 3 additional options making the Label page larger than the standard window size. 

Would you mind splitting Behaviour and Appearance into 2 separate pages? The Label Page is getting quite busy so that would help keeping the pages more balanced and also avoid having Reset buttons changing too many settings at the same time.

Batwam:
Not direclty related to this PR but I'm getting this error in `journactl` which goes away by updating the lines to
```
		visibleFieldsBox.append(firstFieldComboBox);
		visibleFieldsBox.append(secondFieldComboBox);
		visibleFieldsBox.append(lastFieldComboBox);
```
Screenshot from 2023-06-01 19-29-10
`pack_start` had 4 arguments `[pack_start](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Box.html#Gtk.Box.pack_start) (child, expand, fill, padding)` but [Gtk.Box.append](https://docs.gtk.org/gtk4/method.Box.append.html) only had one.

abisammy:
Done!
image

Not too sure why they are showing in journalctl, I removed all the logs from debugging

abisammy:
> Not direclty related to this PR but I'm getting this error in `journactl` which goes away by updating the lines to pref.js
> 
> ```
> 		visibleFieldsBox.append(firstFieldComboBox);
> 		visibleFieldsBox.append(secondFieldComboBox);
> 		visibleFieldsBox.append(lastFieldComboBox);
> ```
> 
> Screenshot from 2023-06-01 19-29-10 `pack_start` had 4 arguments `[pack_start](https://lazka.github.io/pgi-docs/Gtk-3.0/classes/Box.html#Gtk.Box.pack_start) (child, expand, fill, padding)` but [Gtk.Box.append](https://docs.gtk.org/gtk4/method.Box.append.html) only had one.

That error doesn't necessarily have to be fixed if the [new system](https://github.com/Moon-0xff/gnome-mpris-label/pull/58) is implemented. Btw, do you prefer the new or old notations?

Batwam:
> That error doesn't necessarily have to be fixed if the [new system](https://github.com/Moon-0xff/gnome-mpris-label/pull/58) is implemented. Btw, do you prefer the new or old notations?
good point

Moon-0xff:
Sourced prefs.js from `main` to merge cleanly.

I didn't take a good look to the changes in prefs but I wasn't satisfied with the look it had.

Things left to do is to add back the widgets for the other two album options and make a subcategory in panel for 'Icon'.  
This to my opinion gives a cleaner look to the settings. I wasn't able to do it on time but it's the same layout as this screenshot:
image
As the first subcategory of `Panel`

Moon-0xff:
Minus the scale widget, for consistency I think a spin button would be better, at least for now.

Batwam:
The reason I suggested splitting if because with the additional options, it was making the Label page taller than the deault size, showing a scroll bar and hiding some of the options. Also, with so many options in a single page, the Reset button ends up changing too many settings which can be frustrating when only trying to reset some of the settings.

I agree whtat a spin button would be sufficient but would still suggest splitting the Label page in 2 for the reason listed above.

abisammy:
> Minus the scale widget, for consistency I think a spin button would be better, at least for now.

I belive the scale is easier for the user to view. I also made the scale use the current modular system of adding numbers, therefore it can be applied to many other values for consistency.

Moon-0xff:
> I belive the scale is easier for the user to view.

I could agree with that but it looks out of place.

Batwam:
could we please have a version update and/or sync with `main`? This PR now has a lower rev than the version on EGO and gets updated in the background.

Moon-0xff:
I think this branch it's ready to merge. I'm aware the prefs window is a bit longer than before so the GNOME 40+ window workaround might need some adjustment.

Batwam:
Preferences window looks fine on my end now that the option has been moved to the Panel page.

For info, with the default panel size, I get max icon height (before it starts adding horizontal padding) at 65%. If I use the default 60%, the icon is smaller than its max possible size by 1-2px.Is the icon at it maximum height already when you use 60%?

also, it might be worth specifying in the tooltip that by %, we mean "% of the panel height" perhaps?

Moon-0xff:
Max icon height is 65% or very close to it here too. I deliberately choose an smaller value so the icon *aligns* better with the rest of the panel. I would choose 56% but the 4% extra makes a big difference on a 1080p screen.

Moon-0xff:
I tested the changes with the code we used in #12 

On my end the values jump between 2ms and 3ms, sometimes 1ms.  
With the option turned off the values jump between 1ms and 2ms, sometimes 3ms.

Last time I checked the refresh time it jumped around 0-2ms.

So it looks like there's no reason to worry about shell slow downs, at least on normal circumstances.

Here's the diff for testing:
```diff
diff --git a/extension.js b/extension.js
index 410016e..87c8cfb 100644
--- a/extension.js
+++ b/extension.js
@@ -356,6 +356,9 @@ class MprisLabel extends PanelMenu.Button {
 	}
 
 	_refresh() {
+		log("mpris-label ------------------------------------------------------------");
+		let start_time = new Date().getTime();
+
 		const REFRESH_RATE = this.settings.get_int('refresh-rate');
 
 		let prevPlayer = this.player;
@@ -377,6 +380,8 @@ class MprisLabel extends PanelMenu.Button {
 		this._setIcon();
 		this._removeTimeout();
 
+		let end_time = new Date().getTime(); let step = end_time - start_time; log("mpris-label - total cycle time: "+step+"ms");
+
 		this._timeout = Mainloop.timeout_add(REFRESH_RATE, this._refresh.bind(this));
 		return true;
 	}
```

Batwam:
When [testing the Shortwave slowdowns](https://github.com/Moon-0xff/gnome-mpris-label/issues/7#issuecomment-1336120064), I was getting 5-6ms cycles on average, going up to 250ms for Shortwave so this is definitely within the margin of error.

Moon-0xff:
I'm very satisfied with the current state of the branch and I want to merge it now.

Is there anything to change or add?
@abisammy have you reviewed the changes I made? It could be that I moved/refactored your code wrong!

abisammy:
I'll review the changes now...
Feature wise I'm happy with all the changes.

abisammy:
Don't belive I've spotted any code that should or should not be there...
Also all the changes work for my system, therefore I belive this is ready!

Moon-0xff:
I will merge it then! Thanks @abisammy!  
Also thanks @Batwam for aiding with the reviewing process.

Issue #28, opened 2023-02-25
Upcoming Version Roadmap


Batwam:
@Moon-0xff , 
just so I can be clear on what's in/out in terms of features for the upcoming release and any future release when reviewing `main`:
- what features are to be included/excluded from the next release. Is is basically what is currently included so volume control / open player/label filter will only be implemented the following release?
- Are you simply looking to polish what is in the current `main` without additional features?
- Should I expect further commits to be added to `main` as you work through the open PRs?
- If you want to consolidate based on the current fetures added, is it too late to add forward/backward button to the drop down list? it's a pretty easy change and allows to add prev/next to the controls.
- Could we move Controls (click and later scroll) into a separate sheet under Settings? I still think that it would look neater. By the time we (hopefully) add backward/forward buttons + mouse scroll, that's 6 drop downs and the "Panel" tab is already a little busy.
- If you want to hold in implementing some of the other features recently implemented, what features will be pushed to the following release and what's your timeline for that?

No rush if you are still merging, I am just trying to see what to expect.

I am not 100% clear on the objective of the `'g-properties-changed'` implementation as it's only partially refresh-less. The current implementation avoids doing the mpris request on each cycle but still requies the refresh loop for the label and players list. If we (you) wanted to, I feel like it should be possible to remove the refresh loop entirely. Would that be for a different version or you just don't feel that it's necessary? there would be a few ways to go about it but probably worth a separate discussion/PR.

Moon-0xff:
The current `main` is what i want to submit as an update. I'm not looking to add further changes or even polish it (unless it's required or some things we added are too slow, unstable or something )

The next release will at least include the volume controls and will probably be released in a week or two, if it's too close to the GNOME 44 release then wait until we can test the extension on 44.

Currently the controls settings are too few to justify a new settings page (next release then)

I started working in a complete refresh-less implementation (more like a refresh-rate-less implementation), got very far but the extension expects a defined this.player (and probably some other things) in a number of places, this shouldn't be difficult to solve (i hope) but the required changes will create yet more conflicts with the PRs. It's best to do it later.

Moon-0xff:
I didn't make any commits so here's the diff against the latest `main`

```diff
diff --git a/extension.js b/extension.js
index 432b243..145ee38 100644
--- a/extension.js
+++ b/extension.js
@@ -49,9 +49,14 @@ class MprisLabel extends PanelMenu.Button {
 		this.box.add_child(this.label);
 
[...]

PR #60, opened 2023-05-31
Make updating label event based (superseded by #68)

abisammy:
This is event based way of updating the label that can be used.

However it's not fully event based, for example it has to still use the old system refreshing when no players are active, however when there is one active it is fully event based (try logging in the update method)
Have tested:
- Starting with a player active
- Opening another player 

Still need to test a few things such as:
- ~~Closing a player~~
- ~~Closing a player then opening it~~
Tested and it works 🥳

Let me know if you can think of a method to not have to search for players


abisammy:
- Doesn't remove label when media player closed, need to fix later

Moon-0xff:
Great work with this, however I definitely would prefer an implementation based around the methods and object Gjs provides.

I haven't been able to get a result using gjs signals and the other stuff but maybe you can?

Here's some links for research:  
* gjs.guide on signals: https://gjs.guide/guides/gobject/subclassing.html#signals
* gnome-shell source code (signals are implemented everywhere): https://gitlab.gnome.org/GNOME/gnome-shell/-/tree/main/
* My previous attempt to make the extension refreshless (follow the thread): https://github.com/Moon-0xff/gnome-mpris-label/issues/28#issuecomment-1445162694

Moon-0xff:
Signals might be the wrong implementation, I'm not sure, definitely read gjs.guide for good measure.

Batwam:
May I suggest finalising and merging the open PRs before embarking into this exercise considering how much it might be changing the underlying code base?

Moon-0xff:
It might make more sense to merge this one first actually.  
Most of the open PRs benefit greatly from this change.

abisammy:
TODO:
[] Conver to GObject
[] Fix bug where doesn't remove label when media player closed
[] Test and identify more bugs

abisammy:
@Moon-0xff @Batwam Made the changes I need to this banch, I think these changes need to be tested extensively...
Would it be possible for you guys to test this. 

Please note I am aware of the bug that text is not removed when paused.

Moon-0xff:
The diff against `main` is too big.

For this reason I avoided adding extra indentation when converting `MprisLabel` to `GObject` on 28b96aed4fb8db2119f38cca2e10ceba3691d395.

There's also way too much lines with changed indentation/style but no changed code. This further breaks `git diff` and `git blame`

Sorry, but this is just too difficult to review, can you remove the extra indentation, and revert the style changes?

Batwam:
As a side note, I did a bit of research on the topic and found that it is possible to ignore a commit in the blame by creating a file called `.git-blame-ignore-revs` and including the hash of the commit to be ignored. 

Not sure if this is a github specific setting as I haven't tested it but if reformatting is required, this might be worth keeping im mind.

Moon-0xff:
> As a side note, I did a bit of research on the topic and found that it is possible to ignore a commit in the blame by creating a file called .git-blame-ignore-revs and including the hash of the commit to be ignored.

Very useful. I don't think it applies here though!

If it works as I think it works then we can probably do #64 without issues.

Batwam:
good point, I hadn't noticed #64 but that would certainly work there. My understanding is that it should be as simple as adding the file in the root directory but it would be worth checking that it works as intended. the diff will probably still show the changes which is why it would be easier to implement if there is a PR purely dedicated to it rather than combining that with other code change.

abisammy:
Been testing this, couple bugs I noticed:

- [ ] Icon doesnt update when changing size
- [ ] Label doesnt update when changing text
- [ ] Label doesnt change when changing max size
- [ ] Label doesnt change when updating Blacklist & whitelist

Will fix in coming days

abisammy:
Also I just noticed the update method is trigerred when switching applications, perhaps this can be disabled to further optimise performance.

abisammy:
> Please do not add changes that doesn't pertain to the pull request, ideally the changes should be explained only by the PR title

All changes in the PR are related to removing the refresh rate, the previous commits were bug fixes for it?
Ill create a new branch from main anyway...

abisammy:
> Sorry, it looks we aren't on the same page, my problems with the branch are explained in [this comment](https://github.com/Moon-0xff/gnome-mpris-label/pull/60#issuecomment-1589045737).
> 
> I use `git diff` and `git blame` to review and understand the submitted code, this branch doesn't play well with those commands and I can't confidently know where are the changes in the code. Therefore I can't review or merge this.

If you reopen the branch, I can make a commit to remove the indents and styling... This saves me from manually having to copy the changes...

Moon-0xff:
Can you undo the styling changes cleanly? I'm skeptic about it.

Moon-0xff:
Amazing!  
I will review these changes when I'm able. I was expecting a bigger diff honestly.

players.js
```diff
@@ -41,8 +41,14 @@ const dBusInterface = `
 	</interface>
 </node>`
 
-var Players = class Players {
+var Players = 
+GObject.registerClass(
+	{
+		Signals: {'list-changed': {},}
+	},
+class Players {
```
Moon-0xff, line 49: shouldn't extend from `GObject.Object`?

abisammy: True, will fix now

abisammy:
Merged with new changes

abisammy:
New bug:
- [ ] For some reason when opening and closing players, the amount of events infinitley grows when firing events

Moon-0xff:
> Mpris Label: Error: Expected type string for argument 'text' but got type boolean

Odd, perhaps the result value of `removeTextWhenPaused`?

abisammy:
> > Mpris Label: Error: Expected type string for argument 'text' but got type boolean
> 
> Odd, perhaps the result value of `removeTextWhenPaused`?

I had thought so aswell...

I will look into it over the next few days

abisammy:
For some reason the bug isn't being reproduced at the moment...
It could be that I had the wrong branch installed that had the bug.
I'll continue monitoring over the next few days, otherwise the feature might be working as intentionally.

Moon-0xff:
I've been examining the changes of this branch and I've been thinking of rewriting the `Players` class so we don't need to handle the inner workings of `Players` from `extension.js` (which is a workaround, it was not intended)

I forked this branch to do this, I can't show anything useful yet.  
But the general idea is to use `proxy-changed` and a signal when `players.selected` changes to update the label/icon accordingly instead of `list-changed`

Moon-0xff:
Branch is here under the name `refreshless`, I removed a few things, added support for GNOME <42 and made the *structure* changes I wanted.

I started *backwards* to make the structure changes first, and then connected everything. Sadly I wasn't able to make it work, I got this error:
```
JS ERROR: Error: Could not guess unspecified GValue type
update@[...]/players.js:251:8
```
And I don't know how to solve it.

Moon-0xff:
For a clearer log, I've submitted this changes, plus the ones I've added separately as another PR (#68). 
You can, of course, add more changes by creating a PR targeting the `refreshless` branch.

Thanks @abisammy for this PR!

PR #68, opened 2023-06-22
Refreshless Implementation

Moon-0xff:
It looks like GNOME 3.38 for some reason or another isn't able to support a refreshless implementation for this extension, at least not how I've tried to implement it.

The last commits were tested and developed in GNOME 43, the branch should additionally work with GNOME 42 and 44.  
It might or might not work for 40 and 41.

This is a continuation of #60 and it's still WIP.


Batwam:
I've done a quick check and it all seems to be working as expected so far. I played around changing setting to see if it would update and everything I tried worked. Well done!

Moon-0xff:
That's hard to believe! To my experience the player needs to be selected manually, the label doesn't appear until a refresh is triggered, and the icon is sometimes missing.

Batwam:
ah, my bad, it looks like it didn't refresh properly and I was still using `main`... Ok, I can confirm similar issues with the automatic switching not working. icon loading seems ok.

Moon-0xff:
Things not working:
* Auto-switch option
* Hide when paused option

It looks like those aren't easy fixes. @abisammy did solve them if memory serves but I'm thinking in a simpler (but unclear) solution. I want to add those as a separate commit to `main`, first to another branch so `main` isn't broken for so long.

abisammy:
> Things not working:
> * Auto-switch option
> * Hide when paused option
> 
> It looks like those aren't easy fixes. @abisammy did solve them if memory serves but I'm thinking in a simpler (but unclear) solution. I want to add those as a separate commit to `main`, first to another branch so `main` isn't broken for so long.

For the hide when paused the previous solution was a bit 'jank', I dont mind having another go a bit later.

abisammy:
> > Things not working:
> > * Auto-switch option
> > * Hide when paused option
> > 
> > It looks like those aren't easy fixes. @abisammy did solve them if memory serves but I'm thinking in a simpler (but unclear) solution. I want to add those as a separate commit to `main`, first to another branch so `main` isn't broken for so long.
> 
> For the hide when paused the previous solution was a bit 'jank', I dont mind having another go another day.

Moon-0xff:
The general idea is to use a more 'signal based' approach, for a *simpler* code without workarounds. I'm thinking of using timeouts like #17

Moon-0xff:
Removed a repeated comment

PR #69, opened 2023-06-27
Hide the label completely

abisammy:
This implements the changes requested in order to hide the label when changed, mentioned in https://github.com/Moon-0xff/gnome-mpris-label/pull/68.

This branch can also be used to implement the auto switch option...

The difference between this is and the previous solution is that it uses the new optimised event system.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -27,7 +27,6 @@ class MprisLabel extends PanelMenu.Button {
 	_init(){
 		super._init(0.0,'Mpris Label',false);
 
-		this.hideTimeout = null
 		this.settings = ExtensionUtils.getSettings('org.gnome.shell.extensions.mpris-label');
 
 		const EXTENSION_INDEX = this.settings.get_int('extension-index');
--- a/extension.js
+++ b/extension.js
@@ -372,20 +372,19 @@ class MprisLabel extends PanelMenu.Button {
 		this.player.connect('updated', () => {
 			const REMOVE_TEXT_WHEN_PAUSED = this.settings.get_boolean('remove-text-when-paused')
 			const REMOVE_TEXT_PAUSED_DELAY = this.settings.get_int('remove-text-paused-delay')
-	
+
 			if(REMOVE_TEXT_WHEN_PAUSED && this.player.playbackStatus==="Paused"){
-				this._pauseTimeout = setTimeout(()=>this._setText(this.players, true),
+				this._pauseTimeout = setTimeout( () => this._setText(this.players, true),
 				REMOVE_TEXT_PAUSED_DELAY*1000);
-			} else{ 
+			} else{
 				if (this._pauseTimeout != null)
 					clearTimeout(this._pauseTimeout);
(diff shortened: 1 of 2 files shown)


Moon-0xff:
Looks pretty good! I will take a closer look to the code and test it later but at a glance I might accept it without further changes!

Just to nitpick there's an extra newline on 388.

Moon-0xff:
[Remove the timeout in disable](https://extensions.gnome.org/review/38373)  
[Like repositionTimeout](https://extensions.gnome.org/review/38376)

https://gjs.guide/extensions/review-guidelines/review-guidelines.html#remove-main-loop-sources

Moon-0xff:
I paused a player to listen to other, but the button disappeared!

It should display the placeholder string, and keep the button on the panel.

abisammy:
> I paused a player to listen to other, but the button disappeared!
> 
> It should display the placeholder string, and keep the button on the panel.
> 
> Another one:  
> It looks like the player can't be switched.  
> edit: bug already present in `dev-branch`  
> edit2: might be better to solve it in another PR, unless the fix is trivial

The bug where button disappears is trivial, sinply need to set text to placeholder instead!

Will work on it later

Moon-0xff:
Works very well! I will probably rewrite the last commit a bit as it uses javascript *things* that aren't so clear.

Moon-0xff:
Well I can't come up with a better solution on the spot. Any ideas? If not I'm happy to merge this regardless, not a big deal.

abisammy:
> Well I can't come up with a better solution on the spot. Any ideas? If not I'm happy to merge this regardless, not a big deal.

We could split it into three functions...

```
setTextBase(placeholder)
setText() -> calls setLabelBase(false)
setTextPlaceholder()  -> calls setLabelBase(true)
```

Moon-0xff:
Sounds more complicated!

Moon-0xff:
I suppose we could use `clearTimeout` to clear the timeouts on disable and `setTimeout` to create `this._repositionTimeout`

but `setTimeout` is used just once on `gnome-shell` and `clearTimeout` never. I wonder why.  
I also can't find where are they declared, I searched in the gnome-shell repo and gjs. Probably it's a binded C function.

Moon-0xff:
I replaced `setTimeout` with `GLib.timeout...`, it's probably the better option.

Last commit is a partial commit, we still need to replace the call to `clearTimeout`

Moon-0xff:
The last commit message is wrong, I reverted `label.js` to 9afb6190258adc61604a6dae43e113e38f578c6d

Moon-0xff:
I tried to call `hide` and `show` directly but `journalctl` shows a lot of warnings.

There's repeated code when disabling the timeouts, that might be better to solve in another PR.

I think the branch is now ready to be merged. What do you think?

abisammy:
Yep look sgood to me!

Moon-0xff:
Great!  
Thanks again @abisammy !

Issue #67, opened 2023-06-16
Cleaner settings


abisammy:
I think this should wait until we remove compatibility with gnome 40 (https://github.com/Moon-0xff/gnome-mpris-label/issues/62) however I think there are a couple things we can improve in the settings page.
1. As mentioned by @Moon-0xff make it less spread sheet like and add sliders and different form of input (https://github.com/Moon-0xff/gnome-mpris-label/issues/63)
1. Add more interactive blacklist inputs (https://github.com/Moon-0xff/gnome-mpris-label/issues/61#issuecomment-1586093503)
1. Reduce repeteaded code e.g in `prefs.js`:
```js
settings.reset('left-click-action');
settings.reset('middle-click-action');
settings.reset('right-click-action');
settings.reset('scroll-action');
settings.reset('thumb-forward-action');
settings.reset('thumb-backward-action');
settings.reset('volume-control-scheme');
```
4. Reorganise settings page tabs e.g:
- Panel
- Label
- Icon
- Behaviour
- Filters
- Control
(maybe thats too many tabs)
5. Merge some settings, or disable or hide settings if they shouldn't be available (for example if the user disables the source icon, they shouldn't be able to change the settings for the album art, this may confuse them)

I'm happy to work on this, but as mentioned above, waiting until we just have to worry about gnome 40 might be best.

Moon-0xff:
I really don't know what to change or add to the settings, I mostly want to experiment, see what works, what doesn't.

I suppose the best guideline we can refer to is [GNOME's human interface](https://developer.gnome.org/hig/)

Batwam:
Honestly,to modernise the settings panel I'd recommend going straight to libAdwaita. It's pretty neat and, thanks to the various generic function created to add the various widgets, should be relatively easy to update.

It would cut compatibility to gnome 42+ but even Debian Stable is at gnome 43 these days...

See guidance here: https://gjs.guide/extensions/development/preferences.html#preferences-window

Moon-0xff:
It might not be controversial at all, I've been looking at [repology](https://repology.org/project/gnome-shell/versions) and there's few distributions using a version lower than 42.

I think the most prominent one is OpenSUSE Leap stuck at 41.

Batwam:
[...]

PR #73, opened 2023-08-14
Double click feature

michlampert:
Hi!

Sometimes I felt like I was missing buttons in my mouse (or touchpad) to trigger some options, so I decided to implement double click feature, which work similar to this in portable speakers or headphones.

After changes menu will contain additional options:
- switch to enable/disable double-click feature
- field to configure time interval between successive clicks
- three comboboxes with options for double-clicks next  to single-clicks options

Before:
image

After:
image
image


michlampert:
Thanks for your feedback @Moon-0xff!

I hope that changes will meet your expectations :smile: 

Photo of menu for reference:
image

Moon-0xff:
> I hope that changes will meet your expectations 😄

Indeed.

Normally I would prefer to avoid the ternary operator but it looks like a good fit here.

The changes to the menu are good.  
For consistency I would underline "Single click" and "Double click" as "Mouse bindings" is underlined.  
I would also  put a blank space between the button bindings and the scroll binding (there's a commented line that produces a blank line somewhere).

michlampert:
image

Moon-0xff:
Great! I don't have more changes to request. Good job!

I will soon inspect the code more closely and perhaps push to this branch, if isn't a problem on your end (is it?).  

After that I will test it a bit and merge it.

Moon-0xff:
I've added some inconsequential (or should be inconsequential) style changes.  
I will add a few more tomorrow.

Let me know if you have any thoughts about, or if there's problems with my commits.  
And don't be afraid of pushing!

Moon-0xff:
I refactored the code to make it easier to understand (I hope).

The logic should be the same.

Moon-0xff:
Alright, I think I'm done now. Please review the changes and if everything looks good to you I will merge it.

One thing to note, I didn't review on detail the `prefs.js` changes, it looks like it behaves like it should and we are thinking of redoing the whole thing (#67).

michlampert:
Looks good to me :wink: 
I'm looking forward to see it in the next release!

Moon-0xff:
Great! I will merge it then.

Thanks again for the PR @michlampert!

PR #75, opened 2023-09-01
Libadwaita update

Batwam:
Hello,

I have updated the Preferences window to be more aligned to the GTK4 design of apps like gnome-settings using ActionRow and PreferencesGroup. Other changes may be made in the future (use sliders, ...) but for now, I focused on this single design update and maintained the original look to avoid creating too many changes at once. 

**before:**

**after:**

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -65,7 +65,7 @@ function fillPreferencesWindow(window){
 	page.add(group);
 	addSwitch(group,'auto-switch-to-most-recent','Switch to the most recent source automatically:',"This option can be annoying without the use of filter lists");
 	addSwitch(group,'remove-text-when-paused','Hide when paused:',undefined);
-	addSpinButton(group,'remove-text-paused-delay','Hide when paused delay (seconds):',0,10800,undefined);
+	addSpinButton(group,'remove-text-paused-delay','Hide when paused delay (seconds):',0,9999,undefined);
 	addSpinButton(group,'refresh-rate','Refresh rate (milliseconds):',30,3000,undefined);
 	addEntry(group,'label-filtered-list','Filter segments containing:',"Separate entries with commas, special characters will be removed\n\nThe targeted segments are defined in code as:\n\t\A substring enclosed by parentheses, square brackets,\n\t or between the end of the string and a hyphen");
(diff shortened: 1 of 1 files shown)


Batwam:
I'd also like to discuss some minor tweaks for consistency. For example:
- in gnome-settings for example, row descriptions don't have `:` at the end
- I have used subtitles as it's standard in apps like gnome-settings unless the text was too long. For longer text, I automatically add an info button (popover). We can leave as-is or use the info popover whenever there is a tooltip text. Let me know.
- we could potentially have a reset button for each line which would make the interface more busy but allow to only reset individual settings and remove the need for the button at the bottom (and associated code which would be generated automatically within addSwitch,...). I rarely use the reset button as it resets too many settings at once for my liking.

Moon-0xff:
Hello @Batwam!

I'm very pleased with the changes so far, although I will need to inspect them more closely.

As discussed in #67, or at least the *conclusion* I get from it is that we should align with GNOME design, but not be afraid of breaking that style if it fits us.

With that in mind, I agree with point **1** and **2**.

I also agree to add an individual 'reset button' for each setting and to replace the drop-downs. As a separate PR.

I noticed the filters settings has barely changed and it might be better if the changes to it are recorded in another PR, unless the changes are easy to add.

Batwam:
That sounds good. I consider that the update is complete so I'll just issue a final commit to remove the various `:` and that should be ready for review.

Moon-0xff:
As a side note, I tested if using `950` as the default height was problematic with an smaller screen. I down-scaled my monitor to 1280x720 and the window cuts off when it touches the screen border, so no problems there.

Moon-0xff:
I noticed just a few problems:
* Wide entries don't align with the adwaita *style* (as mentioned)
* At minimum window size the triple combo box looks wrong 
* Some repeated lines of code that probably we can skip

I don't have any problems to merge this to `main` as it stands. The flaws are minor and better to be worked in other(s) PR(s).

The only change I'll request is to remove the tab on line 20. But it's such a small thing I think I'll push with the `--author` option. ~~Be ready to be impersonated!~~

PR #76, opened 2023-09-06
add reset buttons in Preferences

Batwam:
See first pass implementation of reset buttons. Note that I had to adjust some of the supporting text to make everything fit as it's getting relatively busy.

The buttons work for all widgets with the exception of the ComboBox which don't update to the default values. To be clear, it does reset gsettings but I can't figure out how to make the button also reset the comboBox value in a generic way which would work in the `buildResetButton()`. Would you mind having a look at it?

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -198,6 +198,12 @@ function addDropDown(group,setting,labelstring,options,labeltooltip){
 		thisResetButton.set_visible(true);
 	});
 
+	//label for debugging
+	let thisLabel = new Gtk.Label({
+		label: 'default: '+thisComboRow._defaultValueIndex.toString()
+	})
+	row.add_suffix(thisLabel);
+
 	row.add_suffix(thisResetButton);
 	row.add_suffix(thisComboRow);
 
--- a/prefs.js
+++ b/prefs.js
@@ -19,14 +19,14 @@ function fillPreferencesWindow(window){
 	page = addPreferencesPage(window,'Panel','computer-symbolic');
 
 	group = addGroup(page,'Icon');
-	let showIconComboBox = addStringComboRow(group,'show-icon','Show source icon',{'off':'','left':'left','right':'right'},undefined);
+	let showIconComboBox = addDropDown(group,'show-icon','Show source icon',{'off':'','left':'left','right':'right'},undefined);
 	addSpinButton(group, 'icon-padding', 'Icon padding', 0, 50, undefined);
 	addSwitch(group, 'symbolic-source-icon', 'Use symbolic source icon', "Uses an icon that follows the shell's color scheme");
 	addSwitch(group,'use-album','Use album art as icon when available',undefined);
 	addSpinButton(group,'album-size','Album art scaling (in %)',20,250,undefined);
 
 	group = addGroup(page,'Position');
-	let extensionPlaceComboBox = addStringComboRow(group,'extension-place','Extension place',{'left':'left','center':'center','right':'right'},undefined);
+	let extensionPlaceComboBox = addDropDown(group,'extension-place','Extension place',{'left':'left','center':'center','right':'right'},undefined);
 	addSpinButton(group,'extension-index','Extension index',0,20,"Set widget location within with respect to other adjacent widgets");
 	addSpinButton(group,'left-padding','Left padding',0,500,undefined);
 	addSpinButton(group,'right-padding','Right padding',0,500,undefined);
@@ -68,7 +68,7 @@ function fillPreferencesWindow(window){
 	let fieldOptions1 = {'artist':'xesam:artist','album':'xesam:album','title':'xesam:title'};
 	let fieldOptions2 = {'artist':'xesam:artist','album':'xesam:album','title':'xesam:title','none':''};
 	let fieldOptions3 = {'artist':'xesam:artist','album':'xesam:album','title':'xesam:title','none':''};
-	let [firstFieldComboBox, secondFieldComboBox, lastFieldComboBox] = addTripleStringComboBox(group,'first-field','second-field','last-field','Visible fields and order',fieldOptions1,fieldOptions2,fieldOptions3,undefined);
+	let [firstFieldComboBox, secondFieldComboBox, lastFieldComboBox] = addTripleStringDropDown(group,'first-field','second-field','last-field','Visible fields and order',fieldOptions1,fieldOptions2,fieldOptions3,undefined);
 
 	//Reset Button
 	addButton(group,'Reset Label settings', () => {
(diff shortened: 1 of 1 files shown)


Moon-0xff:
Looks good but I think it clutters the interface. I would prefer if the button is hidden unless the setting value has been changed, keeping the space where the button is for alignment. No idea how to do that though.

> The buttons work for all widgets with the exception of the ComboBox which don't update to the default values

My bet is that `combobox.set_active_id(settings.get_string(setting))` isn't declared on the *right place*.

With all the refreshless stuff I noticed calls in GJS signals sometimes don't work if declared in *a wrong place/context?* no idea.

Batwam:
I made some changes quickly to see if it could be possible to only show the reset button if the user setting isn't default. The good thing is that when we press `reset`, the key is deleted so rather than compare the current setting Vs default, we can simply check if there is a user defined value or not to decide if we show the reset button or not.

The approach is a little different with the wide entry but I played around with the settings for secondary icons and it seems to work.

Just a rough test for proof of concept but it shoud give an idea. Once again, I focussed on entries/switch/spinbutton.

Also in the todo would be to hide the reset buttons for disabled double click drop downs.

Batwam:
we could also use the rotated icon and put the reset button on the left to avoid the misalignment:
image

Batwam:
it gets busy pretty quickly with multiple drop downs so we might not want to include it for `addDoubleStringComboBox()` or `addTripleStringComboBox()`

Moon-0xff:
> we could also use the rotated icon and put the reset button on the left to avoid the misalignment

Not a bad workaround.

> it gets busy pretty quickly with multiple drop downs so we might not want to include it for addDoubleStringComboBox() or addTripleStringComboBox()

Well the reason we do that is to not lengthen the option list too much.

The triple combo box as it stands takes too much space and I don't like it, but the double one doesn't look bad on it's own.

Perhaps adding the reset button for each row instead of for each option is a good compromise.

Batwam:
> The triple combo box as it stands takes too much space and I don't like it, but the double one doesn't look bad on it's own.

Note that is it currently set at a wider width than necessary. This was done to keep the single combo width the same as the spinbutton on the first page but could be reduced for the double and triple (all 3 are using `buildStringComboBox()` which is where the width is set manually to 105).

Moon-0xff:
I was thinking if we can avoid comboBoxes, they are difficult to work with.

I searched in the docs and it turns out [comboBoxes are deprecated](https://gjs-docs.gnome.org/gtk40/), the recommended alternative is [dropDowns](https://gjs-docs.gnome.org/gtk40~4.0/gtk.dropdown), which sounds like the same thing, but using them might give us less headaches.

Moon-0xff:
> the other things I noticed is that the drop downs in gnome-settings have more of a flat look to them with no borders

Looks a bit out of place. But if it does indeed give us less headaches than comboBoxes then it's a fair tradeoff.

> I found an example here if that helps you

I can definitely see how we could implement after looking at it.  
And if we can hookup the selection to a signal (and the reverse of that) then I think all the headaches are solved. I don't know if this is possible at all, but the hierarchy graph and the methods I found seems to allow such a thing.

Moon-0xff:
I made this by referencing the link you shared and the comboBox function, it *should* be a drop-in replacement for the comboBoxes but it doesn't show a menu when clicked, it does show the selected option (I think) so something it's working at least.

Batwam:
yeah, if you replace `Adw.ComboRow` with `Gtk.DropDown` it shows the content of the list. It's strange that it doesn't work with `Adw.ComboRow` as they are meant to be equivalent and I prefer the look of ComboRow.

I have played around with the code a bit and have something which appears to be working. There are some edge issues (reset button appearing on startup) but generally it works. I'm still keen to keep a generic function to build the DropDown for 1, 2 or 3.

I haven't extended to the multiple drop downs yet but I agree with the idea of the single reset button for all 3 settings. One idea could be to have setting as an array and reset all items in the array in a `forEach` loop perhaps as part of `buildResetButton()`?

Batwam:
ok, that's working pretty nicely for me. Main issue outstanding is the fact that the reset buttons always appear on startup but that shouldn't be too hard to solve.

Moon-0xff:
> I also haven't checked if the reset buttons at the bottom work as intended but since we have reset buttons everywhere now, we might want to remove them altogether?

There is a significant amount of settings for each page, and resetting each one individually would definitely be annoying.

> Main issue outstanding is the fact that the reset buttons always appear on startup but that shouldn't be too hard to solve.

The logic to make the reset buttons appear is flawed, I tried to make them behave correctly but there wasn't a good way of making them work with comboBoxes.

Batwam:
this last commit seems to fix it for the reset buttons on startup. I found `settings.get_user_value()` and `settings.get_default_value()` which greatly help figuring out whether the current settings are the same as default or not :-)

Moon-0xff:
That's great! It doesn't react to the press of the page reset button though.  
To solve that I'm thinking of re-structuring the code that handles the resets, or, make the page reset button virtually push all the individual buttons. Both are valid choices.

Batwam:
yeah, getting the button to press all resets crossed my mind.

Note that I noticed that when change for example a switch and then put it back manually to the default value, the reset button doesn't go away which would be neat. I have implemented a solution for switches but extracting the default values and comparing is a bit tricky as each setting has a different type (integer, boolean, ...).

I have played around with another change which involves hidding the double click settings when the option is disabled in [46b8ce7](https://github.com/Moon-0xff/gnome-mpris-label/pull/76/commits/46b8ce76f7554951007d66f9cbcafcf924f42f5f). What do you think about it?

Moon-0xff:
> I have played around with another change which involves hidding the double click settings when the option is disabled in [46b8ce7](https://github.com/Moon-0xff/gnome-mpris-label/pull/76/commits/46b8ce76f7554951007d66f9cbcafcf924f42f5f). What do you think about it?

Well, the greyed out bindings serves as a prompt to enable the double clicks, which many people will probably want to do.

I suppose this change tries to keep the bindings as a *single widget row*, I guess that because I was trying to do the same.  
My idea was to divide the bindings into two tabs, one for single-click and another for double-click, that would achieve both goals.

Moon-0xff:
> btw, I just discovered the other day that if you are in an app like gnome-center and click on an element, you can then press Ctrl + Shift + i to find useful info like this. You can even change properties on the fly to see what that does

It looks like it can be done in any Gtk application. Including our prefs window.

Moon-0xff:
I tried and failed to store the default index of the dropDowns to trigger the reset painlessly. I will share the broken commit.

Batwam:
I like the idea of saving this property, this is great. I've added a label for debugging and can confirm that the default index is indeed save so that's a start.

Moon-0xff:
> Btw, is it just me or it's not possible to get logs out of pref.js? 

On my end `log( )` works fine. I used it to work out what value was `settings.get_default_value(setting)` returning (a GLib.Variant object).

Batwam:
Just needed to put the dropdown selection within the 'clicked' callback rather than the Reset button creation.

Still need to find a way to get these pesky reset buttons to disappear when the bigger button is clicked! It's not the end of the world and they would go away next time the window is opened but it would still be nice for them to go away...

Moon-0xff:
> would you consider using a similar secondary_icon_name with 'edit-clear-symbolic' icon for addEntry rather than the reset button? or is it better to keep consistent with the reset of the widgets on the page and use the reset button?

I think the context is subtly different. I wouldn't expect both buttons to behave the same way. Though I'm saying this after looking at the question and picture.

I suppose I wouldn't mind the slight inconsistency if it gives us an easier implementation, a considerably bigger text field, or any other advantage.

Moon-0xff:
Last problems I think are better solved by re-structuring prefs. Not something to do on this PR though.

We should start to tidy up the code and get it ready for a merge.

Batwam:
please try this. it shoud clear the rest buttons on reset by checking if the values match the default. I haven't tried for DropDowns yet but the same logic can hopefully apply.

Batwam:
ok, I've implemented the reset button removal when the values match the default. This made the section of code hiding the buttons on click redudant so I removed that and did some general varaible name cleanup for consistency.

I believe that the original objective of the PR is now achieved although I would stil consider using the secondary icon for Gtk.Entry [as mentioned above.](https://github.com/Moon-0xff/gnome-mpris-label/pull/76#issuecomment-1716019237)

Moon-0xff:
> I'm not a big Glib.Variant expert and struggled to extract values from get_default_value the same way for all variant types but found that [with .print(true)](https://gjs.guide/extensions/development/preferences.html#extension-js) we can produce a string with the settings value which can then be used for comparison. [...]

Thanks for the explanation. When I first saw the commit I didn't understand the use of `.print(true)`

> I believe that the original objective of the PR is now achieved although I would stil consider using the secondary icon for Gtk.Entry https://github.com/Moon-0xff/gnome-mpris-label/pull/76#issuecomment-1716019237

I think it's a bit more consistent as it stands.

> Just one other thing but I don't believe that it is direclty related to this PR. Is it me of the "use symbolic icon" option no longer works? I haven't looked at it in detail but I noticed that it wasn't changing when I switched.

Works fine on my end.

Batwam:
Yep ok, for the symbolic icon, it looks like it must be because my Spotify package didn't have the icon as it works with Rhythmbox. All good although we could probably add "(if available)" to the description.

Batwam:
I don't have much else to add to do PR. Any final comments/suggestions?

Moon-0xff:
It might be better to add an explanation on `HACKING.md`.

As for the code I inspected several times now, and while I'm not *satisfied* with it I don't see a problem to it. I think I'll merge this now.

PR #78, opened 2023-09-22
Simplify DropDown functions

Batwam:
I didn't manage to slip that one in before the merge!

Anyway, it wasn't directly related to the reset button anyway. I never felt liked the way we have 3 functions doing essentially the same thing so I thought that it would be a good idea to combine `addDropDown()`, `addDoubleStringDropDown()` and `addTripleStringDropDown()` into a grand unified function. I also simplified the code within the addDropDown to avoid repeated code in the process. All in all, this removed 60+ line of duplicated code...

### DIFF
--- a/prefs.js
+++ b/prefs.js
@@ -210,26 +210,17 @@ function addDropDown(group, settingsList, labelstring, optionsList, labeltooltip
 	return thisDropDownList;
 }
 
-function addSingleDropDown(group, settingsList, labelstring, optionsList, labeltooltip,width){
-	if (!width) //default width
-		width = 105;
-
+function addSingleDropDown(group, settingsList, labelstring, optionsList, labeltooltip,width=105){
 	let thisDropDown = addDropDown(group, settingsList, labelstring, optionsList, labeltooltip,width);
 	return thisDropDown;
 }
 
-function addDoubleDropDown(group, settingsList, labelstring, optionsList, labeltooltip,width){
-	if (!width) //default width
-		width = 135;
-
+function addDoubleDropDown(group, settingsList, labelstring, optionsList, labeltooltip,width=135){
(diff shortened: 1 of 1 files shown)


Moon-0xff:
Indeed this is a great simplification of the code. But I liked the 3 functions as an interface.
The new, unified function requires far more syntax.

Can this code or a derivation of it be wrapped on 3 separate functions with the same syntax (parameters) as before?

Batwam:
ok, I have created 3 separate functions which then call the generic function. The only thing these functions do as this stage is set the default width which avoids having to define it when calling the individual functions.

We could go further and add some more case specific code into each of the three functions but I see some drawbacks to this:
- remove the need to use an array in addSingleDropDown. The array isn't really required so we could go with the simple string but we then have a different format for addSingleDropDown Vs the other two which feels inconsistent
- in `addDoubleDropDown`, assume that the options for both dropDowns are the same. This would be sufficient in our case but would make the function less generic.

Batwam:
Saw the updates. Fine with me 👍

Moon-0xff:
I don't mind the inconsistency between parameters.

> Saw the updates. Fine with me 👍

Great!

Moon-0xff:
I noticed the new functions names aren't the same. But the `String` part isn't helpful. DropDowns aren't well suited for boolean or numeric types.

Other than that I suppose there's not much else to discuss for this PR is it?

Batwam:
That should be it. Please check just in case that the reset button still works and that changing the values does indeed update the settings. If it all works we should be good to go.

Moon-0xff:
No problems on my end as far as I tested.

And in any case I want to push an update with the new changes soon, so it's time to test `main` more thoroughly.

Moon-0xff:
Thanks again @Batwam!

PR #79, opened 2023-09-25
Gnome45 version

Batwam:
See tentative update. I have split this into multiple independant commits with the priority being to make as little change to the underlying code structure in the hope that the gnome45 variant is limited to some line swaps.

Note that for this purpose, I have include 2 initial commits which should apply to the current `main` as I'm relying on them to make the gnome45 version work. everything after these 2 commits is gnome 45 specific.

At this point, the extension loads, appears in the panel with an icon, menu, and preference settings window. What I can't get to work is the import of the buildLabel var/function. For the time being, I have overwritten with with a static text. Would you mind looking into it?

image

PS: just noticed that the Settings window doesn't open from the menu


Moon-0xff:
I don't have an environment with GNOME 45 installed right now, but I will test it when I can.

> Would you mind looking to see how the buildLabel could be imported? I just can't quite figure out what the syntax should be...

Perhaps the `var buildLabel =` part of the function declaration shouldn't be kept
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#standard_import

Batwam:
I installed gnome-boxes as flatpak and installed gnome-nightly from it. 

Importing `Players` works as intended as it's a class. On the other hand, `buildLabel` is a cross between a `var` and `function` but I don't really want to change the original code if we can avoid it...

Batwam:
ok, it's pulling the laebl just fine. Turns out there was a REFRESH_RATE variable left which was crashing the code. I noticed that it has become more sentitive to this sort of thing. Another example was [this error here](https://github.com/Moon-0xff/gnome-mpris-label/pull/79/files#diff-cbd3ad962010bb97aaa7d45e4a08400e6ee51212f9c9b3803aa7d6b513fe90b7R482).

anyway, we have the label now. Still have this issue with the album cover only loading when I turn the extension on/off...
image

Batwam:
ok, the covers seem to work fine on Ubuntu 23.10. 

One thing I did notice however is that the menu items are no longer aligned. It used to look like this
image

Now it looks like this as Ornament is set to `HIDDEN` rather than `NONE` by default.
image

This appears to have been done as a [recent change in gnome](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2843) which changed the default from `NONE` to `HIDDEN` to remove the left padding. We can restore the original look by setting the style back to `NONE` which should still work in the gnome42 version:
image

Batwam:
Would you consider merging the non-gnome45 specific commits into `main` to minimise the changes? These commits should be minor and could be tested now and included in the upcoming release.

Batwam:
Ok, no worries. We still have some time to get the gnome45 version out but Mantic Minotaur is releasing at the end of next month so it would be good to have the gnome version out mid-Oct.

I don't have much to change as I think that what we have in this PR works. There may be a way to avoid putting `settings` back into the pref.js and define it at Global level but I can't figure out how with the new method.

The import method I'm pretty confident is correct. I did notice as I split the imports that we currently seem to be importing libraries which aren't being used but that's probably not a deal breaker.

For getSettings(), the lookupUrl method works. There may be other ways to do it but I doubt that would significantly simplify the code at it's just a one liner currently.

The album cover issue is either a gnome-nightly or rhythmbox specific issue. It would be worth testing this more but the issue doesn't seem to be directly related to the changes in this PR and might have more to do with the issue we we're already experienced with Spotify. Anyway, I'll take the action to test with rhythmbox on Ubuntu 23.10 to see how that looks.

Moon-0xff:
> I was also asking whether you are open to make changes to main which aren't strictly required but help minimise code changes between the gnome42 and gnome45+ versions.

Some of them are to be included because they are fixes of coding errors (that didn't matter because it seems that gjs in 43,44 is more lenient).

The others might not be necessary or beneficial depending on the strategy we will use to handle both versions.

> We still have some time to get the gnome45 version out but Mantic Minotaur is releasing at the end of next month so it would be good to have the gnome version out mid-Oct.

Sounds good.

> There may be a way to avoid putting settings back into the pref.js and define it at Global level but I can't figure out how with the new method.

[Rules don't allow it](https://extensions.gnome.org/review/38147).  
And I just noticed that current `main` is also holding `settings` as global so the new version I submitted to ego will get rejected (oops!).

>I don't have much to change as I think that what we have in this PR works. There may be a way to avoid putting settings back into the pref.js and define it at Global level but I can't figure out how with the new method.

>The import method I'm pretty confident is correct. I did notice as I split the imports that we currently seem to be importing libraries which aren't being used but that's probably not a deal breaker.

>For getSettings(), the lookupUrl method works. There may be other ways to do it but I doubt that would significantly simplify the code at it's just a one liner currently.

👍

> The album cover issue is either a gnome-nightly or rhythmbox specific issue. It would be worth testing this more but the issue doesn't seem to be directly related to the changes in this PR and might have more to do with the issue we we're already experienced with Spotify. Anyway, I'll take the action to test with rhythmbox on Ubuntu 23.10 to see how that looks.

If it works on Ubuntu/G4Music I don't how it could be related to our extension, but gjs being less lenient and crashing in some places that didn't before could mean that there's broken functionality deep into the code.

Moon-0xff:
Certainly we will need to test the initial 45 version a lot more before releasing it.

Batwam:
Ok, I guess you can cherry-pick the first 2 commits of this PR then?

For info, I'm in discussions with the Just Perfection guy who is suggesting some changes so I'll probably end up starting this gnome 45 port from scratch to keep things clean but at least we know that it can be made to work.

Batwam:
PR closed pending `main` code update to support gnome45 transition

Issue #80, opened 2023-09-26
Single dropdown rows not resetting


Moon-0xff:
The **page reset button** doesn't reset the single dropdown rows, it does reset the setting itself, but the dropdown doesn't return to the default value.

This bug also exists on 810445d78c5e43387d7bcc1b6e431ac85d4282e4 so it wasn't introduced on 7d9b6d67535b3bb959eca6e2f65ae7e09a7eb9d2.  
Double dropdowns and triple dropdowns reset just fine.

Moon-0xff:
Fixed on #81

PR #81, opened 2023-09-27
fix reset for single DropDown

Batwam:
in relation to #80, see update to fix reset button for single DropDowns. The variables should have been arrays.


Moon-0xff:
Ah! Now I get it. `addDropDownList` returns an array.

Another fix would be to return the first element of the array on `addDropDown`:
```js
function addDropDown(group,settings,labelstring,options,labeltooltip,width=105){
        return addDropDownsList(group, [settings], labelstring, [options], labeltooltip,width)[0];
}
```

But I think your solution is better.

I'll merge this right away.

PR #82, opened 2023-09-28
Code cleanup for future compatibility

Batwam:
Hi, I would like to propose some "transparent" updates to the code which don't impact the way it works or looks but will fix some upcoming issues ahead of the transition to gnome 45 based on some the errors I had to fix to get the gnome 45 version.

In the order of the commits, this would be:
- propose to use strict to enforce some "discipline" there
- fix undefined variables (pref.js)
- settings defined as global variable in pref.js
- remove unused dependancies (players.js getIcon() )
- remove unused variables (players.js)
- remove unused function inputs (players.js)
- add schema info to metadata.json
- changes which will become requirements for the gnome45 fork but are transparent for gnome 42 (schema definition, popupMenu Ornament)

The idea would be to merge this and only then introduce the gnome45 PR based on this cleaned up code.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -14,7 +14,7 @@ const { buildLabel } = CurrentExtension.imports.label;
 let indicator = null;
 
 function enable(){
-	indicator = new MprisLabel();
+	indicator = new MprisLabel(ExtensionUtils.getSettings());
 }
 
 function disable(){
@@ -26,10 +26,10 @@ function disable(){
 var MprisLabel = GObject.registerClass(
 	{ GTypeName: 'MprisLabel' },
 class MprisLabel extends PanelMenu.Button {
-	_init(){
+	_init(settings){
 		super._init(0.0,'Mpris Label',false);
 
-		this.settings = ExtensionUtils.getSettings('org.gnome.shell.extensions.mpris-label');
+		this.settings = settings;
 
 		const EXTENSION_INDEX = this.settings.get_int('extension-index');
 		const EXTENSION_PLACE = this.settings.get_string('extension-place');
--- a/extension.js
+++ b/extension.js
@@ -47,7 +47,7 @@ class MprisLabel extends PanelMenu.Button {
 		});
 		this.box.add_child(this.label);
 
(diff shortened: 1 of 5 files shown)


Moon-0xff:
I would separate 1fe61d4053dacd06bf1fa5e3c735892dc437e3cf and try to merge it first, is almost granted that ego will reject the next version until `settings` is de-globalized.

Batwam:
yep, that's the main one and likely to be a must-have like you said and you can cherry-pick it if you want. However, the other ones are very minor if you want to have a look at them also.

Moon-0xff:
Well, I did.
It also merged c885407f41e6a05f3fc356167ba0962e144e46c7 for some reason, but that's fine.

Batwam:
Ok, that's good. I've synced the branch so there is very little left in term of diff. Feel free to review and merge when you find the time. I can reimplement the gnome45 port and resubmit a cleaned up PR which will hopefully be relatively short too.

Batwam:
ok, those were the main changes. if you look at the diff now, these are minor updates fixing basic/obvious coding errors like undefined variables.

I am now pushing some updates to replace the use of ExtensionUtils as dependancy injection. This was recommended in my chat Just Perfection and will also help limit the code differences between gnome42 and gnome45 versions. Again, no change on the look and feel.

Batwam:
ok I was told to use dependency injection everywhere and avoid Global as much as possible in label.js so I have updated the code accordingly:
image

Moon-0xff:
Looks great, I don't have a problem with merging this as it stands. Though I'm worried `use strict` could silently break stuff.

Batwam:
Yeah, I did notice that with 'use strict', unless the issue is in the main code, it wasn't getting any error, it just fails and I had to go dig to find where the issue came from. It looks like gnome45 will flag these errors or refuse to run anyway so I'm ok if you want to revert that commit.

Batwam:
I guess we should change the variables in label.js to lowercase before merging for consistency but I was running out of time...

PR #83, opened 2023-10-01
More cleanup for future compatibility

Batwam:
Hello, I noticed a couple more minor changes I'd like to make which apply to the gnome 43/44 version too. It's not 100% necessary to include them in `main` as they can always be part of the gnome45 diff but I though it might be cleaner to merge those first to minimise the differences between the two versions. 

The gnome45 version is now ready to go by the way, I'll publish once this is confirmed.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -387,7 +387,10 @@ class MprisLabel extends PanelMenu.Button {
 		}
 
 	//settings shortcut:
-		this.menu.addAction(_('Settings'), () => ExtensionUtils.openPrefs());
+		let settingsMenuItem = new PopupMenu.PopupMenuItem('Switch Automatically');
+		settingsMenuItem.setOrnament(PopupMenu.Ornament.NONE); //to force item horizontal alignment
+		settingsMenuItem.connect('activate', () => ExtensionUtils.openPrefs() );
+		this.menu.addMenuItem(settingsMenuItem);
 	}
 
 	_refresh() {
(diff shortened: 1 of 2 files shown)


Moon-0xff:
I'll wait a bit in case there's something else to add.

Also, please check your email inbox.

Batwam:
Invitation accepted 👍

The commits in this PR were actually already in the previous gnome45 merge request. This is mostly aimed at keeping the gnome 45 PR to the strict minimum but considering the location and nature of the changes, we can certainly keep them within the gnome45 branch/commit/diff.

Moon-0xff:
> Invitation accepted 👍

Thanks!

> The gnome45 port is ready to go so I'm not expecting further changes

Alright, then I'll merge this now. I agree it's better to keep the differences between `main` and the 45 version small.

If the 45 diff is small enough then I think a patch file would be more manageable, though it will require a script and/or some user guidance when installing manually.  
I think it's better than keeping a branch that will inevitably drift from `main`.

Batwam:
@Moon-0xff Sincere apologies as I just noticed a mistake in the last commit due to copy-paste. Could you please update change 'Switch Automatically' with 'Settings' here: https://github.com/Moon-0xff/gnome-mpris-label/pull/83/files#diff-19a9a3efb94bf1b84229bda71ab0ef5badcdbf34ba3a0c9e4152a82c569a21e6R390

Moon-0xff:
Done.

PR #84, opened 2023-10-02
code update for gnome45 compatibility

Batwam:
As promised, updated code for gnome45 compatibility ready for testing. It has been designed to require as little code change as possible to minimise the risk of conflict later on.

Regarding how we "manage" the 2 versions going forward, we could code from the gnome 43/44 version but update the install script to automatically pick up the Gnome version and apply a patch if gnome >44 at the time of install? Obviously, that's just for the github version, 2 sets of files will be uploaded to EGO. Another alternative would be for the gnome45 version to be `main` and the older versions to be where we apply the (reversed) diff. It all depends on what we consider should be the "standard" gnome version for most users.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -1,24 +1,29 @@
-const Main = imports.ui.main;
-const PanelMenu = imports.ui.panelMenu;
-const PopupMenu = imports.ui.popupMenu;
-const {Clutter,Gio,GLib,GObject,St} = imports.gi;
-const ExtensionUtils = imports.misc.extensionUtils;
-const CurrentExtension = ExtensionUtils.getCurrentExtension();
-const Volume = imports.ui.status.volume;
+import * as Main from 'resource:///org/gnome/shell/ui/main.js';
+import * as PanelMenu from 'resource:///org/gnome/shell/ui/panelMenu.js';
+import * as PopupMenu from 'resource:///org/gnome/shell/ui/popupMenu.js';
+import Clutter from 'gi://Clutter';
+import Gio from 'gi://Gio';
+import GLib from 'gi://GLib';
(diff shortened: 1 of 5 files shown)


Moon-0xff:
I expect an easier time managing the differences by using the 43/44 version as base. I think the diff is small enough to be served as a patch file so I'm going on that route. I will push an experimental branch soon enough.

Moon-0xff:
Done. See the new branch `45-as-patch`.

What it's left to do is an explanation on README and perhaps your idea of updating the install script to the detect if the host is running 45. Though I think doing that automatically might not be a good thing.

Perhaps just a simple message like:  
`If you are running GNOME 45 you will need to patch the compatibility diff file [instructions here], then run this script again`

Batwam:
Yeah, I think that the patch would work. I have generated a diff based on this PR which we can use if you are happy with the content. We could also update the install script to apply the patch if the version is >44? That would help me as I'll be switching to gnome45 later this month.

Moon-0xff:
I think we commented at the same time 😅. See comment above.

Moon-0xff:
I think the index serves the same purpose as the SHA in the commits.

And I know the patch files will work until `patch` can no longer apply it, same as when merging branches on git.

> We will probably need to update the revisions to add gnome46, etc... in the future but we will be able to update the diff file manually for that right?

Yes, provided that 46 will be compatible with 45, it should be just a matter of updating the patch.

Moon-0xff:
Alright so this PR was technically merged through d43bacd46c5c600b303fd6a7cd8650a422a375e8.

As I said before, it needs an explanation on the README and at least a mention on `install.sh`.  
It also needs to be tested.

I'll close this now.

PR #85, opened 2023-10-04
Apply gnome45 patch automatically

Batwam:
This code checks for the gnome version and applies the patch automatically as long as `patch` is installed (unlikely but wasn't installed on the default gnome-nightly so I though I might as well include the check...). The code applies the changes to the copied files in the target folder rather than the base files so we don't get code changes in the git folder itself.

Let me know if that works on your end.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,8 +10,8 @@ cd "$(dirname "$0")"
 printf "\e[32mCopying extension files to target directory:\n\e[0m"
 cp -Rv ./* $DEFAULT_INSTALL_DIR
 
-gnomeVersion=$(gnome-shell --version | awk -F'[ .]' '{print $3}') #get major Gnome version
-if [ $gnomeVersion -ge 45 ]; then #apply patch if Gnome 45+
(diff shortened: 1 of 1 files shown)


Moon-0xff:
Patching was likely to fail for a lot of other reasons. I also used short-circuiting to save a couple of lines.

Batwam:
Yeah, I agree, good to capture potential patch failures in the future as we change other parts of the code. I'll be on gnome45 soon enough so I will hopefully be able to capture patch issues when testing.

Batwam:
One other thing is that the "Using git" method which copies the files directly in to the extensions folder will not work for Gnome 45 as the files won't be patched. We might need to remove it or mark it as "Gnome 43/44" only?

Moon-0xff:
I wrote the 45 patch instructions to work for both methods of installing the extension ("Using git" and "Using git and the installation script").

Though now the installation script (if all goes well) will patch the extension so we should rewrite that part of the README.

Moon-0xff:
Merging this.

Issue #57, opened 2023-05-26
Hide the extension's panel when nothing is playing


vixalien:
Currently, a small blob appears when nothing is playing. Would be cool to hide the extension completely when nothing is playing... like the Removable media extension.

![image](https://github.com/Moon-0xff/gnome-mpris-label/assets/37999241/a640029d-02ae-4f0d-90f4-e454b5de8dc5)

Moon-0xff:
First of all, are you hovering the pointer over the space?

Moon-0xff:
Yes, hiding that blob was something I've been wanting to do since the release of GNOME 40 but never tried!

I will keep it on the top of the ~~imaginary~~ to-do list.

vixalien:
haha take your time...

Batwam:
I guess we are already removing the container and putting it back in when relocating the extension so we already have the building blocks within `_updateTrayPosition()`? 

Just need to remove `this.container` if the label is blank and trigger `_updateTrayPosition()` when the text comes back in?

nkcong206:
I see in Aylur's Widgets extension they can do it but it makes lag. I hope you can take a look at it.

Moon-0xff:
Yes, It uses the same functions I was thinking of, `PanelMenu.Button.hide` and `.show`.

I think the lag comes from spawning the panel media buttons, we have the same problem in #39

Batwam:
for what it's worth, i found out that if we include `this.set_style("border: 0px;");` within `_init()`, the panel-button won't shot any highlight when the mouse is over it. Controls (scroll, right click, ....) still work. This means that when the label is in "hide when paused" mode, there is virtually no way to tell it's there. Right click will still work so the menu can still be accessed. All in all, it makes the extension integrated like info in the panel rather than a button but perhaps makes it less obvious that there are funcitonality accesible on click.

Alternatively, it would probably be possible to set the border to 0 when pause and then back to 1px when the label is showing some text.

![image](https://github.com/Moon-0xff/gnome-mpris-label/assets/56710655/ccf9727c-cd7c-416b-9215-0134be532f68)

vixalien:
I think instead of that, straight up hiding/showing the panel icon would be better if possible.

Batwam:
Yeah, it's a bit of a hack. I just wanted to out it out there as I stumbled upon this behaviour while working on another PR and there seemed to be some lags generated with using show/hide.

abisammy:
[...]

PR #87, opened 2023-10-28
Hide extension when no playing

Batwam:
Proposed implementation for #57 
image

Note that I have made the following design assumptions to be dicussed:
1. I have kept the code to hide if no player is active but this will only trigger is the "hide extension" option is selected. We can change this easily if we want to continue hiding the extension when no player is open by default.
2. I have used the same delay mechanism to hide the label and extension. Upon testing it felt weird to see the label disappear right away when I would have expected the label to stay as long as the pause delay isn't reached. Technically we could have different settings but I think that this would be overkill.
3. the code only checks the active player so, if the active player is paused but another one is playing, the extension will still hide itself. (this will not happen if the user has Auto Switch enabled)
4. This option is disabled by default


Moon-0xff:
This one is too aggressive. Might be good for some users?

In any case the code somewhat conflicts with the 'hide when paused' code on `label.js`. I think we should bring that code outside of `label.js`, manage everything on the widget side.

Moon-0xff:
I'll try to make my own solution on the next few days, closing this for the meantime.

PR #90, opened 2023-12-01
Mute Spotify during Ads

Batwam:
Hello, it's me again!

I came across this [Spotify Ad-Block extension](https://github.com/danigm/spotify-ad-blocker) and found that it would be easy to implement something similar as we already have all the building blocks to identify mpris info, mute/unmute, ... Essentially, the idea is to spot if `trackID` contains `spotify/ad/` which was actually suggested by a user.

Let me know what you think. I have tested it for a while and it's working well so far. It won't block chromium based ads but despite my investigations, I don't think that this will be possible as `chromium` appears in the streams list but not in the `mpris` list so there is no way to know if the browser it being used for an ad or not unfortunately. Note that Spotify has a [tendency to spawn multiple streams](https://community.spotify.com/t5/Desktop-Linux/Spotify-app-floods-music-streams-list/m-p/5676192#M22317) when ads are started, that's why I'm muting all of them.

I included a default lag of 1400m (essentially 5x300ms refresh cycles based on default refresh rate) to avoid unmuting before the end of the ad. If you use the Spotify app, please check if that default duration works for you. When muting, the icon/album will show the mute icon so people can visually see why it has gone quite. I could have gone of an OSB but I thought that the icon change was more subtle.

Finally, if the user disables the extension in the middle of an ad, extension will unmute as part of `disable()`. I know that this is quite unlikely but I felt that the extension should leave the system as it would be if it wasn't there. I made the code generic as possible to allow the ad-blocking functionality to be extended to other sources of possible in the future.

image


Moon-0xff:
If we can read `this.activated` from spotify-ad-blocker, then we can change the logo or implement any other behaviour when an ad plays, without having to duplicate the underlying logic.

I definitely see value on adding that.

Moon-0xff:
Yes, I don't justify the effort of implementing it either, but as I said, there's value on adding it.

Of course it will be prone to breakage but it looks like the codebase is fairly stable, so I don't think it would be a problem to maintain.

But I'm pretty sure it will break eventually, like if the extension stops being maintained.

So overall: Nah. If you want to develop it I would accept it but there's not much value on that work honestly.

PR #92, opened 2024-01-23
Change track on scroll option

### LINKED ISSUE #91: Add change track functionality to scroll
While side buttons is pretty comfortable to use in this scenario, I myself is more used to quick changing tracks by simply scrolling through. I can work on this feature myself if you're willing to accept my PR.

RozeFound:
Closes #91

I added added a bit of buffer before changing the track, so only the last one of scroll events calls the function


Batwam:
Thanks for sharing this. I think that you will also need to remove the timeout in `_disable()` as the EOG reviewers usually expects everything to get cleaned up on exit.

Regarding the timeout mechanism, I understand that it will execute 100ms after the last scroll event as the timeout is getting reset. This might make it feel a bit sluggish. Do you think that it would be possible to execute the action right away, and instead use the timeout purely as some sort of lock which prevents the song skip action from being repeated as long as it's not complete? 

Regarding the duration of that lock/timeout, I would have thought that 1000ms would be reasonable. If you wanted to skip multiple songs by scrolling, what sort of frequency would feel natural to you? I tried to see how quickly I would tap and got to 70bpm/1110ms personally.

RozeFound:
> Do you think that it would be possible to execute the action right away, and instead use the timeout purely as some sort of lock which prevents the song skip action from being repeated as long as it's not complete?

Unfortunately I have no idea how to implement it in any other way, I simply can't find the way to determine if that was last scroll or not without the delay. Timeout here works purely as a buffer, if you try to scroll in this 100ms gap, the timer will reset, otherwise it'll execute the action. I think 100ms is barely noticeable anyway with average human reaction time around 200-250ms.

RozeFound:
I reduced the delay to 75ms, now it feels almost instant, but still prevents from rapid song-skipping. And you can skip songs relatively fast if you want to skip multiple.

Batwam:
Timestamps might be the easiest solution for this. Something simple like this using timestamps works:
```
	case "track-change":
		let timeGap = Date.now() - this.__track_change_tid
		let scrollDelay=700;
		if ( ! this.__track_change_tid || timeGap > scrollDelay){ //700ms gap to prevent repeated skips
			this.__track_change_tid = new Date().getTime();
			this._changeTrack(delta);
		}
		break;
```
it will execute right away but will not allow another skip for another 1000ms. What do you think? we use timestamps in other parts of the scope (see `players.js`).

I don't think that you need to define `this.__scroll_delay` globally as it's only used in one function. I think that the way you had it previously  was clear enough.

RozeFound:
To sum up

My solution:
- Any length
- Fixed speed
- Delayed execution

You solution:
- Fixed length
- Any speed
- No delay

No idea what's better tbh

RozeFound:
I have an idea, we can combine two solutions

We change the track, create timestamps, and if the user is trying to scroll in this gap, we'll reset timestamp.

```js
	case "track-change":
		let timeGap = Date.now() - this.__track_change_tid;
		let scrollDelay=100;
		if (!this.__track_change_tid || timeGap > scrollDelay) {
			if (delta > 0) 
				this._activateAction("next-track");
			else if (delta < 0) 
				this._activateAction("prev-track");
		}
		this.__track_change_tid = new Date().getTime();
		break;
```

This way we have basically my solution but without delay

Batwam:
yeah, the updated solution with timestamps works. I actually reworked the solution and tried to implement it using your original timeouts in the meantime which appears to work. I still feel like timestamps are easier to work with. See below for what it's worth:
```js
	case "track-change":
		if (this.__track_change_tid)
			GLib.Source.remove(this.__track_change_tid);
		else {
			if (delta > 0) 
				this._activateAction("next-track");
			else if (delta < 0) 
				this._activateAction("prev-track");
		}

		this.__track_change_tid = GLib.timeout_add(GLib.PRIORITY_HIGH,
			this.__scroll_delay, ()=> {this.__track_change_tid = null;});
		break;
```

RozeFound:
Glib timeouts like this gives unnecessary overhead I think especially with high priority
Anyway, how does current solution looks?

Batwam:
that seems to work fine. I think that in the spirit of the rest of the rest extension where everything is customisable and very little is hard coded, we should probably allow the user to vary the scroll delay just in case. Would you mind adding a spin button to this section and set your preferred values as default?
image

you will need to update the xml file in schemas. to generate the `gschemas.compiled`, run `glib-compile-schemas schemas/` within the extensions' folder. [more details here](https://gjs.guide/extensions/development/preferences.html#creating-a-schema) if needed.

Batwam:
see proposed implementation in the commit above. Let me know what you think.
image

RozeFound:
How about we make it grayed out when track change is not selected? Or even hide entirely

image

Batwam:
looking good, I just updated from visible to sensitive for consistency with `Enable double clicks` on the same settings page

Batwam:
That's all looking good to me. Is the extension working the way you wanted now?

@Moon-0xff if no further comment, I'll go ahead and merge.

RozeFound:
Yep, works like I wanted it to

Moon-0xff:
This is good! Only I would have squashed the merge instead of adding the whole commit history.

I'm also thinking that this code can be refactored to allow execution of all or most actions (probably some will crash the shell though).

Issue #93, opened 2024-01-30
label color change


slaygun:
hey can u lmk how i could change the color of the label text which gets displayed

Batwam:
hi,

we don't currently have an option for this but you can update `extensions.js` manually around row 485 as follows (example here with red font):
```js
	_setText() {
		try{
			if(this.player == null || undefined)
				this.label.set_text("")
			else {
				this.label.set_text(buildLabel(this.players,this.settings));
				this.label.set_style('color: #FF0000');
			}
		}
		catch(err){
			log("Mpris Label: " + err);
			this.label.set_text("");
		}
	}
```
![image](https://github.com/Moon-0xff/gnome-mpris-label/assets/56710655/ea2ebfce-0f66-4590-879b-3850f072e96b)

slaygun:
omg thank u so much im not well versed with javascript so i was having a pretty hard time

slaygun:
![image_2024-01-31_152846243](https://github.com/Moon-0xff/gnome-mpris-label/assets/73815160/23786220-fb87-4c4b-ae9f-57159ace3875)
it works!

Batwam:
No problem, it's a good suggestion and I was actually thinking this weekend about allowing to change things like the font. 

I'll see if I can integrate an option in the Preferences to allow the user to pick a colour easily.

Batwam:
@slaygun could you please test the solution in #94 ?

slaygun:
@Batwam what is the command to get that branch

Batwam:
Sorry, git clone the repo, then get the branch by doing `git checkout font_colour`

slaygun:
![image](https://github.com/Moon-0xff/gnome-mpris-label/assets/73815160/4f386555-df1f-4f48-b339-4f796b95afc6)
yes its working!

Batwam:
ok, that's great. This should allow you to try different colours without having to manually change the code and without having to restart the extension all the time. Let me see if I can integrate a button to select a colour more easily...
![color-button](https://github.com/Moon-0xff/gnome-mpris-label/assets/56710655/a42fa3d2-6af6-4f7a-bf33-223cda25d2b8)

Batwam:
@slaygun I have updated the Preferences to include a button to select the colour. Would you be able to download the latest code and try it?
![image](https://github.com/Moon-0xff/gnome-mpris-label/assets/56710655/6b91d266-5a15-48e5-8abf-774eb9d6d961)

[...]

PR #94, opened 2024-01-31
option to specify label font color

Batwam:
implement solution for issue #93 to specify label font color

image

image


Batwam:
updated to include a color choose widget (and use rgb value in the background): 
image

Moon-0xff:
I think this option shouldn't have a default value (or have a *null* value as default), and the code should be executed only if the user has picked something.  
Font colors are dependent on the shell theme, and our default might blend with the background.

Batwam:
I agree with the fact that assuming that white is the default can be dangerous if people have themes. The main challenge I had is that, while it's probably possible to figure it out, I don't know how to find out what that default colour is. This said, I could leave the default setting blank. This will have the effect to set_style('color:""') which seems to set the colour back to the default. 

The Setting page then looks like this with a transparent colour (since the colour isn't recognised) which indicates that the default font isn't being overwritten): 
image

Is that good/clear enough? If not, we'd need to find a way to extract the default colour so it can be applied to the ColorButton when the setting is blank.

Batwam:
ok, I found how to get the font colour using [this post](https://discourse.gnome.org/t/how-to-get-data-from-extension-js-file-to-prefs-js/15095). As far as I can tell, it's not possible to import `Main` in prefs.js (which was also the issue the person posting had) so I followed the same solution and created a setting which is used to store this value. This setting is being populated/refreshed on started and whenever the color  is getting changed. 

The sole reason I'm doing this is so I can set the colorPicker's colour correctly in the preferences window to match the theme's font when the font color is being reset. Otherwise, the previous solution with the checkered rectangle works as well doesn't necessitate the additional code.

Batwam:
I have reverted the previous commit. it was possible to find the font colour but this added too much complexity to the code and potential hedge cases when trying to keep `addColorPicker()` generic for future uses, depending on whether an alternative key exists which needs to be checked or not, effect of the theme changing after the Settings window is being opened, ...

schemas/org.gnome.shell.extensions.mpris-label.gschema.xml
```diff
@@ -30,6 +30,14 @@
...
     </key>
 
+    <key name="font-color" type="s">
+      <default>""</default>
+    </key>
+
+    <key name="font-color-default" type="s">
+      <default>""</default>
+    </key>
+
```
Moon-0xff, line 40: Unused, leftover from [e6ddde9](https://github.com/Moon-0xff/gnome-mpris-label/pull/94/commits/e6ddde929cbbeec92a8f32c342dee58ffb8f8f3e) and merged to `main`.

prefs.js
```diff
@@ -46,6 +46,7 @@ function fillPreferencesWindow(window){
 
 	group = addGroup(page,'Appearance');
 	addSpinButton(settings,group,'max-string-length','Max string length (each field)',1,150,undefined);
+	addColorPicker(settings, group, 'font-color', 'Label Font color', "Specify alternative for color for label\nDefault falls back to color defined by user theme");
```
Moon-0xff, line 49: First line (`Specify alternative...`) of the tooltip could be dropped.

PR #96, opened 2024-02-20
Code cleanup following font-color option merge

Batwam:
@Moon-0xff  code cleaning minor updates as requested:
- removal of unused leftover key
- shortening of label font color tooltip which now fits as a subtitle

image

note that I have not updated the revision. I would recommend to use the same number (v27) for both gnome 42/43 and 45 version since the content/features are identical.


Moon-0xff:
> note that I have not updated the revision. I would recommend to use the same number (v27) for both gnome 42/43 and 45 version since the content/features are identical.

Version will be assigned by EGO. Bumping version on our own could interfere with updates.

Issue #99, opened 2024-04-12
unknown symbolic icons in label page settings ...


dasnoopy:
[very minor issue) ...as per attached screenshot , in label page setting there are some  unknown symbcolic icons (see filter segments containing and button placeholder settings) - [ extension version  30 - gnome 46 - default adwaita icon theme)

![Screenshot from 2024-04-12 06-46-50](https://github.com/Moon-0xff/gnome-mpris-label/assets/296922/8bd95474-4ae2-4e7a-a08b-3b9e39338a8b)

I'm not a developer but I find out that should be involved this part of code in **pref.js** file.
I replaced the `info-symbolic` icon name with  e.g. `dialog-information-symbolic` and a correct icon appear near the label..

```
function buildInfoButton(labeltooltip){
	let thisInfoButton = new Gtk.MenuButton({
		valign: Gtk.Align.CENTER,
		icon_name: 'info-symbolic',
		visible: true
	});
```

thanks
Andrea

Moon-0xff:
Thanks for reporting this!  
I don't run 46 myself so I'll have to boot up a VM to test (it could be that your system is missing some files!).

Batwam:
Yeah, I have already investigated this as I noticed a similar issue in Gnome nightly. Unfortunately, this icon is not part of Adwaita but of gnome-control-center.

See gitlab issue raised: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/issues/283#note_2027550

Do you have gnome control center installed?

When discussing the issue with others, the recommendation was to package the icon as part of the extensions.

Moon-0xff:
I was wondering where the icon was coming from.  
I'm assuming gnome-control-center is or should be packaged alongside gnome.  
Does gnome work well without gnome-control-center?

If not then I think this is the result of a broken package or misconfiguration.

If using gnome without gnome-control-center is a valid use case scenario, then I think we should test if the icon is available, if not, replace it with an icon that's explicitly packaged with gnome, or inside a package that gnome depends on (regardless of distro).

Moon-0xff:
> perhaps that could be a fallback (pretty sure there is a "fallback" method in St.Icon. Would that be a good solution?

In that case it would be a simple one line diff:

```diff
diff --git a/prefs.js b/prefs.js
index 1ee7752..f8c735e 100644
--- a/prefs.js
+++ b/prefs.js
[...]

PR #101, opened 2024-04-29
Update info button icon to use Adwaita icon 'dialog-information-symbolic'

Batwam:
As #99, our current implementation relies on an icon from Settings (gnome-control-center) which may not be available if the software isn't present or in heavily sandboxed systems. Proposed minor update to info button icon in Preferences window due to currently using a non-Adwaita icon.

Successfully Tested in Gnome 45 and gnome-nightly.
Yaru (Ubuntu default) for which the icon interestingly still looks like an i
image

Adwaita/gnome-nightly
image

Please check with older versions to make sure this icon already existed.


Moon-0xff:
I'll update the gnome45 patch with this instead of merging this PR.

PR #109, opened 2024-10-15
Fix depreciated Clutter.Color for Gnome 47+

Batwam:
Hello,

I noticed in Gnome47 that while everything seemed to work fine, the Menu wasn't showing anymore. Upon inspection, it appears that `determineColors()` was getting blocked by the following updated from Clutter to Cogl: https://gjs.guide/extensions/upgrading/gnome-shell-47.html#clutter-color

See proposed changes applying extra patch on top of the Gnome 45+ patch.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -13,7 +13,11 @@ cp -Rv ./* $DEFAULT_INSTALL_DIR
 GNOME_VERSION=$(gnome-shell --version | awk -F'[ .]' '{print $3}') #get major Gnome version
 if [ $GNOME_VERSION -ge 45 ]; then #apply patch if Gnome 45+
 	echo "You are running GNOME 45 or above. The script will try to patch compatibility for this version. See README.md for details."
-	patch -d $DEFAULT_INSTALL_DIR < patches/gnome45-compatibility.patch && echo "Patch applied!" || echo "Patch failed!" && exit 1
+	patch -d $DEFAULT_INSTALL_DIR < patches/gnome45-compatibility.patch && echo "Gnome 45+ Patch applied!" || { echo "Patch failed!"; exit 1 ; }
+	if [ $GNOME_VERSION -ge 47 ]; then #apply patch if Gnome 45+
+		echo "You are running GNOME 47 or above."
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -13,7 +13,11 @@ cp -Rv ./* $DEFAULT_INSTALL_DIR
 GNOME_VERSION=$(gnome-shell --version | awk -F'[ .]' '{print $3}') #get major Gnome version
 if [ $GNOME_VERSION -ge 45 ]; then #apply patch if Gnome 45+
 	echo "You are running GNOME 45 or above. The script will try to patch compatibility for this version. See README.md for details."
-	patch -d $DEFAULT_INSTALL_DIR < patches/gnome45-compatibility.patch && echo "Patch applied!" || echo "Patch failed!" && exit 1
+	patch -d $DEFAULT_INSTALL_DIR < patches/gnome45-compatibility.patch && echo "Gnome 45+ Patch applied!" || { echo "Patch failed!"; exit 1 ; }
```
Moon-0xff, line 16: Changing `&&` to `{ ... }` doesn't alter behaviour right?  
If that's the case I do agree that the logic becomes a little bit clearer.

patches/gnome47-compatibility.patch
Moon-0xff, line 1: Adding a new patch file is a perfect way of solving this 👍

Moon-0xff:
I don't see any problem or have anything else to add to this.
Thanks again @Batwam!

install.sh
Batwam: Sorry, didn't see the comment. Actually, it does change the way the actions are grouped. With the original implementation, the code was exiting regardless of true/false which meant that it couldn't proceed to the next patch.
- True -> ok message
- False -> failure message
- exit

By grouping using {}, exit only happens with False so:
- True -> ok message -> continue
- False -> failure message -> exit

Batwam: example here, you will see that example #3 doesn't exit as exit it only linked to failure:
image

Issue #48, opened 2023-04-18
play/pause not working with Rhythmbox when playing radio stations


sebo505:
Have set right mouse klick to toggle play/pause.
Works with Rhythmbox when playing podcast.
Works with Rhythmbox when playing music from local library.
Works NOT with Rhythmbox when playing radio streams.

Have noticed that Rhythmbox itself shows stop button instead of pause button when playing radio.

Any way to make toggle play/pause(stop) working in this case never the less?

![Bildschirmfoto vom 2023-04-18 22-00-00--#bug_#Gnome_Extension_#Mpris_Label](https://user-images.githubusercontent.com/65250211/232893255-254156a0-63af-4c79-a488-297b3a2b3e39.png)

sebo505:
Sounds great. Much appreciated !!!

Moon-0xff:
Change committed, Indeed it was an easy and straight forward fix.

If you want to get the fix early you can follow the [instructions on the readme](https://github.com/Moon-0xff/gnome-mpris-label/blob/main/README.md) to install the development version.

Moon-0xff:
New version is out, with the fix for this included.
Thanks for raising this issue!

PR #110, opened 2024-10-17
plexamp play-pause workaround

Batwam:
I recently started to play with Plex and when installing Plexamp, I noticed that while I could Pause, I couldn't resume the songs. the rest of the funcitonalities works (artwork, song info, ...) so that's good.

on further inspection, he reason PlayPause is flakey is because Plexamp incorrectly sets CanPause to false when the song is paused (see below, CanPause changing from true to false):
image

If your widget checks for CanPlay && CanPause at the same time to trigger PlayPauseRemote, it will assume that is can’t control the song. Normally, players leave CanPlay and CanPause to true when these functionalities are enabled which is [the way the specification intends it to be](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:CanPause) so it’s a bug.

I have reported this bug [on the Plexamp forum](https://forums.plex.tv/t/mpris-implementation-flakey/887376/3). However, since it's proprietary, I doubt the Plex team is going to fix it anytime soon so I took it upon myself to include a workaround. Basically, we need to only check CanPlay if the song status is Paused.

Note that technically, we could remove `<method name="Play" />` and use `this.proxy.PlayPauseRemote()`. The main reason I'm using Play is because technically, CanPause is `false` and using Play is a fit clearer in my view. We could change that if you find this overly pedantic.

### DIFF
--- a/players.js
+++ b/players.js
@@ -3,6 +3,7 @@ const {Gio,Shell,St} = imports.gi;
 const mprisInterface = `
 <node>
 	<interface name="org.mpris.MediaPlayer2.Player">
+		<method name="Play" />
 		<method name="PlayPause" />
 		<method name="Next" />
 		<method name="Previous" />
@@ -263,6 +264,11 @@ class Player {
 	toggleStatus() {
 		if (this.proxy.CanPlay && this.proxy.CanPause){
 			this.proxy.PlayPauseRemote();
+			return
+		}
+		//workaround for some players (Plexamp) which incorrectly change CanPause set to false when paused
+		if (this.proxy.CanPlay && this.proxy.PlaybackStatus == "Paused"){ 
+			this.proxy.PlayRemote();
 			return
 		}
 		if (this.proxy.PlaybackStatus == "Playing"){ // fallback to "Stop"
--- a/players.js
+++ b/players.js
@@ -265,10 +265,12 @@ class Player {
 
 		this.proxy.PlayPauseRemote();
 
-		if (this.proxy.PlaybackStatus == "Playing" && prevStatus == "Playing"){ //fallback to "Stop"
-			this.proxy.StopRemote();
-			return
-		}
+		setTimeout(() => { //wait 1s s PlaybackStatus update has a lag
+			if (this.proxy.PlaybackStatus == "Playing" && prevStatus == "Playing"){ //fallback to "Stop"
+				this.proxy.StopRemote();
+				return
+			}
+		}, 1000);
 	}
 	goNext(){
 		if (this.proxy.CanGoNext)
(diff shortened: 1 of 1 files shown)


Moon-0xff:
As the commit states I haven't tested it.

Batwam:
I just gave it a got but it didn't work as it was stopping the song every time I wanted to pause. I'm not sure if there is a lag or something but when pressing, even after the pause action is triggered, `this.proxy.PlaybackStatus` is showing `Playing`:
image

Moon-0xff:
Makes total sense than the application can't switch status fast enough.

I guess we could add a small timer to the `Stop` workaround, but that's quite complex for such a simple action... But it seems to me like the best way to avoid buggy MPRIS interfaces, which aren't rare.

Batwam:
yeah, or we keep things even simpler and just run `this.proxy.PlayPauseRemote()`. Do we really want to stop the song on click considering that this means that the user won't be able to resume and will lose where he/she was in the song?

I have pushed a proof concept with a timer. Based on my pc, 50ms is sufficient to get an updated `PlaybackStatus`, 10ms is too short. I put 1s to be safe as this is done asynchronously but to be frank, I think that it adds complexity and could potential trigger if the machine was extra slow.

Moon-0xff:
The `Stop` workaround was added because rhythmbox doesn't respond to `Pause` signals when using the radio (#48).  
Though adding the workaround with a timeout, specially such a long timeout will probably bring more trouble than it solves.

I guess it's best to revert the changes to https://github.com/Moon-0xff/gnome-mpris-label/pull/110/commits/98f025ee2f6765af424bedf0dfb74ee1c050bc5a

Batwam:
Yeah, I think so too. Would you be able to revert and merge? I'm afraid I might mess it up.

Moon-0xff:
For the record:
```
git revert HEAD~2..HEAD --no-commit
git commit
```
(without the `--no-commit` flag it becomes two separate commits).

Moon-0xff:
It's a little bit frustrating to have two distinct workarounds on such a simple (or supposed to be simple) function.  
But well, Software Development!

A one-size-fits-all solution against any buggy mpris interface sounds to me now, well, impossible.  
I guess we are doing it well adding workarounds (or not adding them) in a case by case basis,  
but that's work I personally would prefer to avoid.

Batwam:
yeah, it's a bit frustrating having to work around improper implementation but that's probably a necessary evil. On the plus side all these players support the same standard and most of it works so it's not too bad. At least this isn't affecting performance for players which properly implement is since is should exist  with the first `this.proxy.CanPlay && this.proxy.CanPause` check.

Issue #89, opened 2023-11-24
Suggestion: Modify the scrolling control volume direction


WhoChangedMyCode:
In the "Controls" of the "Scroll up/down", the current situation is that scrolling up is to lower the volume, scrolling down is to raise the volume, I think this is not in line with the intuitive operation, especially when I use the touchpad of the notebook, the result of the volume control is always the reverse of my first intuitive operation, maybe I am a different kind of bar.

I hope to have some improvement:
Scrolling up is to increase the volume, scrolling down is to decrease the volume. Or add a new option for the user to choose whether scrolling up raises the volume or lowers it.

Moon-0xff:
I also noticed the touchpad gesture to raise/lower volume is reversed. But this is consistent with the global volume widget, and I think the rest of GNOME.

It's a simple change to add though, on `extension.js`:
```js
			case 'volume-up':
				this._changeVolume(1); /* CHANGE THIS TO -1 */
				break;
			case 'volume-down':
				this._changeVolume(-1); /* CHANGE THIS TO 1 */
				break;
			case 'volume-mute':
				this._changeVolume(0);
				break;
		}
	}

	_changeVolume(delta){ /* MULTIPLY delta BY -1 */
		let stream = [];
		stream[0] = this.volumeControl.get_default_sink();
		let streamName = 'System Volume (Global)';

		const CONTROL_SCHEME = this.settings.get_string('volume-control-scheme');
```

Batwam:
agreed with keeping the scroll volume control consistent between the extension and the global volume widget. I just checked and changed from "Traditional" to "Natural" and they both inverted so that's the correct behaviour in my view.

Moon-0xff:
Alright. I will close this as unplanned.

As mentioned: You can modify the source code to do this (just follow the comments I left on my previous comment).

Issue #47, opened 2023-04-03
Incompatibility with Dash-to-Panel Scroll Action


Batwam:
Hello,

I just noticed as I was testing with multiple workspaces that the scroll action on the label was also trigerring the workspaces to switch. When I turned of Dash-to-Panel, this went away so this is not an issue without the extension. I then looked into the settings and noticed that by default, scroll will `Switch workspace`
![image](https://user-images.githubusercontent.com/56710655/229516781-e8980c9d-8419-40e2-807d-903f16e21e9c.png)

I checked and we do have `return Clutter.EVENT_STOP;` in `_onScroll(event)` but clearly both extensions are getting the signal. 

This issue can be disabled which is what I've done but I just wanted to put this out there so we are aware of this limitation.

Moon-0xff:
Should we close this one? Both extensions have options to disable the scroll actions, ours since 34829624994af461f4b1f773ba78160488606307.

We could try to prevent the signal from reaching both extensions (somehow). But ultimately I don't think we should implement complicated stuff for the weird edge cases.

Batwam:
Yeah, it's a known issue but I don't think that there is much we can do about. We could perhaps include a know in the Readme to state that it's a know limitation?

PR #113, opened 2024-11-23
if natural-scroll is enabled, invert scroll inputs

### LINKED ISSUE #112: Invert scroll direction if "natural scolling" is used
Scrolling on a laptop touchpad can use "Traditional" or "Natural" mode:
image

The "natural" mode works quite well for scrolling _content_, e.g. webpages, especially when super-precise smooth scrolling is enabled.

However, it also inverts the scroll direction for this extension, which means that the volume increases when you pull down your fingers, which doesn't make much sense.

Complicating the issue: I also often use a mouse with my laptop, and I don't want that to be inverted! It already works correctly. So ideally the logic is:
- Check if the scroll event is coming from a touchpad or a mouse. If coming from a mouse, use current behavior.
- If coming from a touchpad, check if natural scrolling is enabled. If yes, invert the scroll direction. Otherwise use current behavior.

The traditional/natural scrolling can presumably be read from wherever GNOME stores that setting. I'm not sure how to detect if the scroll input is coming from a touchpad or mouse. I spent some time trying to find documentation on the scroll-event signal but my lack of experience here was showing. There may be some metadata on event that we can use, but I wasn't able to find any conclusive docs on this.

I'd be glad to put in some more effort on making the code change here given some pointer from a maintainer.

cg505:
When natural scrolling setting is enabled, the scroll inputs are essentially the opposite of what's actually physically happening. This is fine for scrolling down a page, but feels very weird for things like volume control.

Invert natural scrolling inputs so scrolling up actually increases the volume, and vice versa.

Fixes #112.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -33,6 +33,9 @@ class MprisLabel extends PanelMenu.Button {
 		const EXTENSION_PLACE = this.settings.get_string('extension-place');
 		const REPOSITION_DELAY = this.settings.get_int('reposition-delay');
 
+		this.mouseSettings = new Gio.Settings({ schema: 'org.gnome.desktop.peripherals.mouse' });
+		this.touchpadSettings = new Gio.Settings({ schema: 'org.gnome.desktop.peripherals.touchpad' });
+
 		this.box = new St.BoxLayout({
 			x_align: Clutter.ActorAlign.FILL
 		});
@@ -171,9 +174,6 @@ class MprisLabel extends PanelMenu.Button {
 		if(SCROLL_ACTION == 'none')
 			return Clutter.EVENT_STOP
 
-		if (event.is_pointer_emulated())
-			return Clutter.EVENT_PROPAGATE;
-
 		let delta = 0;
 		const time_delta = Date.now() - this.last_scroll;
 		switch (event.get_scroll_direction()) {
@@ -196,6 +196,22 @@ class MprisLabel extends PanelMenu.Button {
 
 			default: //exit (do nothing)
 				return Clutter.EVENT_PROPAGATE;
+		}
+
+		// If natural-scroll is enabled, the scroll event we get is inverted.
+		// We need to invert it back.
+		switch (event.get_source_device().get_device_type()) {
+			case Clutter.InputDeviceType.POINTER_DEVICE:
+				// mouse
+				if(this.mouseSettings.get_boolean('natural-scroll'))
+					delta = -delta;
+				break;
+
+			case Clutter.InputDeviceType.TOUCHPAD_DEVICE:
+				// touchpad
+				if(this.touchpadSettings.get_boolean('natural-scroll'))
+					delta = -delta;
+				break;
 		}
 
 		if(!delta == 0)


extension.js
```diff
@@ -171,9 +174,6 @@ class MprisLabel extends PanelMenu.Button {
 		if(SCROLL_ACTION == 'none')
 			return Clutter.EVENT_STOP
 
-		if (event.is_pointer_emulated())
-			return Clutter.EVENT_PROPAGATE;
-
```
cg505, line 176: I'm not sure why this was initially added. It seems like my touchpad has a lot of scroll events that are considered emulated but otherwise look normal. This was causing problems for me because I have another extension with scroll event handling on the entire menu bar. A bunch of scroll events would hit this and propagate through, even though I just wanted to interact with the mpris label.
So that's why I removed it. Could also just make it return EVENT_STOP instead.

Moon-0xff: > I'm not sure why this was initially added.

I don't remember, though [gnome-shell's code does the same](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/slider.js?ref_type=heads#L156).

> This was causing problems for me because I have another extension with scroll event handling on the entire menu bar.

We found the same issue you are describing with Dash-to-panel (#47).

> So that's why I removed it. Could also just make it return EVENT_STOP instead.

I'm not sure if we want "emulated scroll events" to trigger the scrolling action, this would probably change the volume increasing/decreasing speed in some hardware (can you test if this is the case in yours?). The idea is to follow gnome's volume widget on ~behaviour, though we are changing scroll direction here...~ speed, so I'm more inclined to change EVENT_PROPAGATE to EVENT_STOP.

Moon-0xff:
While this might be out of scope, it makes sense that if we decide to *fix* the direction of the volume slider, we should do the same for the built-in global slider. This can indeed be done, though I'm not sure how.  
I *believe* the volume widget is somewhere inside `Main.panel` and can be referenced (`Main.panel...`), if we can grab the volume slider object, then perhaps we can redefine the [scroll function](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/slider.js#L152).

extension.js
cg505: I found this bugzilla thread: https://bugzilla.gnome.org/show_bug.cgi?id=687573, and these commits: https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/7d4e14f38455977ffb103acd85239ed6b1a9d816, https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/c9741ae3d5f661b2aee352c50ce21fab16694fbe (the last one is somewhat confusing to me since I definitely observe events with `is_pointer_emulated()` true).
It seems that some events are essentially duplicated, one emulated and one not emulated... I don't have the historical reasoning for why this is the case but my impression is that it has to do with retrofitting touchpad support into X11.

I suppose ideally other extensions should also be checking this, but in practice it doesn't make a lot of sense to propagate the events instead of just ignoring them completely.

Upshot: happy to use EVENT_STOP, makes sense.

cg505:
> While this might be out of scope

IMO, it is out of scope of the extension, but definitely this PR.

Options:
1. Monkeypatch volume control (and maybe other sliders like brightness?) in the settings panel.
    - Note: I think if we're gonna do this, we need a solution that works for all the sliders, not just volume. Seems quite confusing if the volume slider works one way but the brightness slider works differently.
    - It's pretty unintuitive that installing this extension would modify the behavior of random UI elements that already exist in the shell, even if it sort of makes sense why we would want that.
2. Try to get GNOME to fix this specifically for the panel sliders by using a similar technique to here. (Uninvert scroll events using awareness of natural scrolling setting value.)
3. Get GTK/Clutter/whoever to fix this upstream so that we can somehow differentiate the physical scroll direction from the intended content movement direction. (Ideal, but complicated and requires much deeper changes.)
4. Do nothing.

I'm inclined to do 4 because I'm lazy. Not really a fan of 1 for the reasons mentioned. 2 and 3 outside the scope of the project.

Moon-0xff:
I do believe it would be of value to also flip the volume slider, and just the volume slider,  
Though, I've tried before of modifying elements of GNOME this way, and while it's possible, it's not straightforward, and it sometimes breaks things in unexpected ways.

So option 4 it is!

Moon-0xff:
Now, this change seems controversial, and also some people would prefer it the other way around as in (#89). So I think this work should be adapted to fit *all kinds of bar*.

I mean, being able to choose between:
* GNOME's default scrolling direction
* Reversed direction if "Natural scrolling" is enabled
* Always reverse the scrolling direction

As reference, this [commit](https://github.com/Moon-0xff/gnome-mpris-label/commit/a9331fc37f0a3be045ff801a275c18e55382f056), `prefs.js` and the schema file. and [gjs guide](https://gjs.guide/extensions/development/preferences.html#creating-the-schema).  
You need to write the schema, compile it and add the function calls where you seem they fit.  
I'll do this if you are done with the PR. Great work!

cg505:
> Always reverse the scrolling direction

Although I see that some people would want the behavior to be consistent with the GNOME behavior, it doesn't seem likely that anyone would want this "always". That is, even if it is inconsistently with GNOME behavior.

I think it should be sufficient to just wrap the code I added in an if statement and add a setting called "respect natural scrolling" or something.

Moon-0xff:
> just ... add a setting called "respect natural scrolling"

Thing is, I also felt the direction reversed when testing it for the first time using the "Traditional" setting, so making this change accessible to "Traditional" users makes sense to me.  
On a UI/UX point of view though, a single boolean setting for this is way better.

Also, sorry for not having a follow-up on this yet.

cg505:
Sorry I've kind of dropped this. Haven't forgotten about it, just been busy. (After all, I still experience this bug nearly daily lol.)

I actually didn't realize that you can scroll on the volume icon in the normal system panel, I thought you meant the bar in the expanded panel. That makes it significantly more important to make it consistent between these two in my eyes.

Issue #88, opened 2023-11-17
Transferring this repository


Moon-0xff:
I would like if @Batwam accepts my proposal of becoming the new owner and maintainer of the repository.

I stopped giving this project much attention and even though there hasn't been new PRs or issues I feel guilty about it.

I don't think this change would affect this project much, @Batwam has done most of the heavy lifting here.

Of course, the decision lies with @Batwam. But I raise this as an issue as a sort of explanation for my neglect, if it starts to show.

Batwam:
@Moon-0xff thanks for raising this. To be honest, I'm quite happy with the way things are working as it allows to cast another set of eyes on new code before it gets released. I believe that this has been beneficial to the stability of the extension and ultimately end user.

If you are wondering if the current pace is causing me any frustration, I can tell you that it's not. In my view, the extension is pretty much feature-complete. While I have some ideas for features which could be integrated (e.g. Spotify ad-block), I am not expecting significant code changes and there is no rush for any new features to be implemented as far as I know.

Moon-0xff:
I'm relieved to know I haven't caused you any frustration.

And I could simply ask you to keep an eye on the repo, push what you see fit to `main`, while I maintain my position as owner and push the updates. That would keep the extension running smoothly.

But I don't think there's a reason why I should keep my position as owner/maintainer instead of relaying it.  
It would better reflect reality if you assume the position of owner/maintainer at that point.  
The checking step to weed out bugs and mistakes could be done as a friendly query.

PR #121, opened 2025-03-24
Update to Gnome 48

MiMillieuh:
Works perfectly fine without any other changes on Gnome 48

### DIFF
--- a/patches/gnome47-compatibility.patch
+++ b/patches/gnome47-compatibility.patch
@@ -34,7 +34,7 @@ index 1f0d6c8..e29d0f9 100644
 -    "version": 38,
 +    "version": 37,
 -    "shell-version": [ "45", "46" ],
-+    "shell-version": [ "47" ],
++    "shell-version": [ "47", "48" ],
      "url": "https://github.com/Moon-0xff/gnome-mpris-label",
      "settings-schema": "org.gnome.shell.extensions.mpris-label"
  }


Batwam:
Thanks for that. I haven't checked gnome 48 yet. Are you sure that all functions work, including the window focus/unfocus "open app" one? I just had a look at the [port guide](https://gjs.guide/extensions/upgrading/gnome-shell-48.html#meta) and there seems to be changes to some functions currently in use like this one https://github.com/Moon-0xff/gnome-mpris-label/blob/8b29641514e0e89d5b6e26e80d2fbaa5862a41d0/players.js#L327

image

MiMillieuh:
Hello, 

Yeah my bad, focussing the window doesn't seem to work... I don't use that function at all so I haven't noticed to be fair.

I've tested the other shortcuts and they seem to all work, the settings app works too and the display is fine.

So basically there's just the window focus that's not working properly according to my testings.

Batwam:
Seems like we might need an extra gnome48 specific patch and update of the script.sh file then.

Would you be able to check if replacing `global.get_window_actors()` with `global.compositor.get_window_actors()` in the line above of `player.js` is sufficient to make the open app function work?

Joris-Peeters:
Hi, I tried this. I edited `~/.local/share/gnome-shell/extensions/mprisLabel@moon-0xff.github.com/players.js` and rebooted, but the app still won't open.



I don't have experience with GNOME extensions, but if there's anything else I can do to help debug this, let me know!

Kerrung:
> apologies as I haven't been able to investigate yet. I tried to install Fedora RawHide in a VM but gnome-boxes no longer seems to work for me... so I might have to wait until Ubuntu 2025.05 is out.
> 
> In the meantime, seems ok to manually set `gsettings set org.gnome.shell disable-extension-version-validation "true"` to allow the extension in the meantime.

@Batwam Hi! Also you can test your extension on GNOME OS with always the latest GNOME Nightly.
https://os.gnome.org
So you no need to wait for the latest Ubuntu release. And if GNOME Boxes have not worked you, you can try Oracle Virtualbox.

CCHunter90:
I really don't think a single feature not working is a reason good enough for the extension to instead be totally unavailable for GNOME 48. The rest of users will continue to have the extension as is, only the ones with GNOME 48 will loose some functionality until that is fixed but at least releasing this would make them have some use of the extension.

Batwam:
@CCHunter90 yeah, unfortunately both the maintainer and myself are very time poor so this extension is mostly unmaintained as discussed in #88.

I just tried in Fedora rawhide and it all seems to work fine as far as I can tell. I did some research and found another extension which also didn't change it's code but included a comment saying that this might break in the future: https://github.com/lucaswerkmeister/activate-window-by-title/blob/main/extension.js#L109-L111

@MiMillieuh for the sake of avoiding adding yet another patch file, I'd suggest adding a similar comment in the code of player.js just before `let windows_list = global.get_window_actors()` and we can ask the maintainer to push an update.

Issue #137, opened 2026-02-27
gnome-shell crashed when changing Position -> Extension place from right to left


Johann-Tree:
I use Ubuntu 24.04.4 LTS and installed your wonderful extension in version 38. After I changed the setting _Position -> Extension place_ 
from _right_ to _left_ gnome-shell crashed:

```
Feb 14 20:58:29 host gnome-shell[3009]: Object St.Bin (0x639d1a6c79f0), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                           == Stack trace for context 0x639d161e90c0 ==
                                           #0   639d162b14a8 i   file:///home/user/.local/share/gnome-shell/extensions/mprisLabel@moon-0xff.github.com/extension.js:122 (2416da508470 @ 129)
                                           #1   639d162b1418 i   resource:///org/gnome/shell/ui/init.js:21 (2e687f70bf0 @ 48)
Feb 14 20:58:29 host gnome-shell[3009]: Object St.Bin (0x639d1a6c79f0), has been already disposed — impossible to access it. This might be caused by the object having been destroyed from C code using something such as destroy(), dispose(), or remove() vfuncs.
                                           == Stack trace for context 0x639d161e90c0 ==
                                           #0   639d162b14a8 i   file:///home/user/.local/share/gnome-shell/extensions/mprisLabel@moon-0xff.github.com/extension.js:126 (2416da508470 @ 246)
                                           #1   639d162b1418 i   resource:///org/gnome/shell/ui/init.js:21 (2e687f70bf0 @ 48)
Feb 14 20:58:29 host gnome-shell[3009]: GNOME Shell crashed with signal 11
Feb 14 20:58:29 host gnome-shell[3009]: == Stack trace for context 0x639d161e90c0 ==
Feb 14 20:58:29 host gnome-shell[3009]: #0   639d162b14a8 i   file:///home/user/.local/share/gnome-shell/extensions/mprisLabel@moon-0xff.github.com/extension.js:126 (2416da508470 @ 246)
Feb 14 20:58:29 host gnome-shell[3009]: #1   639d162b1418 i   resource:///org/gnome/shell/ui/init.js:21 (2e687f70bf0 @ 48)
```

Batwam:
Thanks for raising this. this must be a regression as I never had this issue before but I'm experiencing it now too.

[...]

PR #139, opened 2026-03-03
Update position crash fix

Batwam:
proposed fix for #137 

I included a check with `return` which allows to avoid trying to delete something which isn't accessible... The St.Bin warning is still obscure and will continue to appear in the journal as it's being checked by the `if` statement.

I also took the opportunity to simplify the relocation logic by removing/reinserting the widget rather than trying to relocate it.

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -63,8 +63,9 @@ class MprisLabel extends PanelMenu.Button {
 
 		this.settings.connect('changed::left-padding',this._onPaddingChanged.bind(this));
 		this.settings.connect('changed::right-padding',this._onPaddingChanged.bind(this));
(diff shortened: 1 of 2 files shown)


Batwam:
ok, previous solution worked to avoid the crash but still having the warning showing up bothered me deeply so I dug a bit more. My understanding is that the issue is triggered by the `'changed::extension-index'` calling `_updateTrayPosition()` multiple time in parrallel. this is the reason the warning is issued as there are literally two functions deleting and recreating the widget at the same time. This explains why the exact same function didn't crash on startup.

I changed the logic so that the signal triggers a flag for the widget to be repositioned at the next `refresh()` cycle. This way, the signal can trigger multiple times, the function will only get called ones ~300ms later (default refresh time)

